PR #26, opened 2016-05-24
shivit-chat initial commit

Technoblazed:
Allows the user of both upper and lowercase chat triggers. Found it easier to create a new plugin to carry this out rather than use a previous version.

### DIFF
--- a/scripting/shavit-chat.sp
+++ b/scripting/shavit-chat.sp
@@ -1,0 +1,66 @@
+#pragma semicolon 1
+
+#include <sourcemod>
+#include <sdktools>
+#include <shavit>
+
+public Plugin:myinfo =
+{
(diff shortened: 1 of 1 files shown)


shavitush:
Pretty cool!

Instead of a new module, I think that it would be much better to simply revamp the current ["!R terminator"](https://github.com/shavitush/bhoptimer/blob/master/scripting/shavit-misc.sp#L370-L381) so it does the same thing.

Technoblazed:
Let me get you that now :D

shavitush:
Also, you could replace the first character of the `char[] sArgs` string with `sm_` and fake execute the command as it is instead - that would be neater as you won't need to use a function that turns the whole string into lowercase.

You might also want to send a new pull request from a different branch instead of `master`, AFAIK pull requesting from `master` can create issues with the repository's structure!

PR #28, opened 2016-05-24
Misc |- Uppercase Chat Triggers

Technoblazed:
Updated everything you suggested, works fine on my test server and changed to 1.7 syntax.

### DIFF
--- a/scripting/shavit-misc.sp
+++ b/scripting/shavit-misc.sp
@@ -367,17 +367,17 @@ public Action OnClientSayCommand(int client, const char[] command, const char[] 
 		return Plugin_Handled;
 	}
 
-	// if capitalized !R = run sm_r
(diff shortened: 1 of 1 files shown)


shavitush:
The commit seems to override the whole file for some reason!

And the code is a bit redundant, because a user may for example, type `!wR` and it will not work, as you're only checking for the first character after the chat trigger. 

[I've implemented it by myself for now](https://github.com/shavitush/bhoptimer/commit/fd1851137909f91d96376b7da77ed1a923f14598?diff=unified), thanks for contributing though! I also saw that you've started implementing different gravity and speeds as properties, I'm looking forward for them!

shavitush:
Also, the code you used will run for example, `sm_!Wr` into a console instead of `sm_Wr` (unless I'm wrong)

PR #29, opened 2016-05-26
HSW Additions

Technoblazed:
Code created by https://forums.alliedmods.net/member.php?u=270493 but I'd rather you push it as soon as possible so that I can keep working on the new styles.

- Third Person
- Slowmotion
- Gravity

Should be coming later today/tomorrow.

Thanks


shavitush:
I expected another kind of implementation, sick though!

shavitush:
Thanks for the contribution!

PR #30, opened 2016-05-26
Removal of Radar on player spawn

Technoblazed:


shavitush:
Could you do the following so it matches the rest of the code?

1. Change `1<<12` to `(1 << 12)` - shouldn't be a `#define` and it would be just fine to use `(1 << 12)` in-line and if possible, add a comment after this line and say it's going to hide the radar.
2. Below `SetEntProp()`, add `return Plugin_Stop;` and remove the extra newline under this block.

Technoblazed:
Also, did you have a chance to look at the test css I've been reproducing for the webpanel? [Here](http://moderation.joshog.com/bhop/index.php)

shavitush:
I did and it looks great! Feel free to send a pull request for it if you want but just make sure to add it as a setting in the config!

PR #31, opened 2016-05-27
Customized Red Webpanel

Technoblazed:
Includes config option to add a new red themes webpanel, including
slight changes to the code to make the default template appear with a
better layout.


shavitush:
Thanks for the contribution!

PR #37, opened 2016-06-01
Allow keys on ground & block prejump in start zone

OfficialExtan:
It's better if the player can use all keys on ground if he selects a style like w-only. Makes walking on ground easier!

Also it would be better to block prejump on start zone because every prejump is different and the player doesn't start with the same speed every time. With this fix, times can be compared much better. :) You could also do a cvar for this.


Technoblazed:
+1 for this, but would probably want a cvar for the two :3

shavitush:
I don't agree about blocking jumps in start zone, sorry.
If you add a cvar for it though, I'll gladly merge the pull

OfficialExtan:
Added cvar for this :)

PR #39, opened 2016-06-02
Fix Noclip Abuse

OfficialExtan:
This is significant:
If player uses noclip and flies 45 - 90 degrees up in the air in the start zone and disables noclip before he leaves the start zone, he can start the timer and fly/shortcut right into another stage or even to the end zone. (Bug found randomly by CRNO)
This line of code fixes it and is working fine on my server :)

Could be also fixed differently e.g: Setting the Y-speed of the player to 0 when he disables noclip in the start zone :+1:


shavitush:
sweet, thanks!

PR #45, opened 2016-06-07
Fix noclip prespeed abuse

OfficialExtan:
So I found out a way to fix the noclip prespeed abuse.
I looked into the code of the InsideZone function and saw that you move the y-origin 5 units up. This means if the player touches the upper corner of the start zone with his feet, his timer will start AND his speed will remain. So i copied this function and just called it IsInsideZone and removed the playerPos[2] += 5.0; and added another condition of the prespeed part.
Well it works and fixes it but looks like it can be made easier and shorter. Maybe you know how.
Also i tested to just remove the playersPos line in the InsideZone function but this will result in starting the timer even if the player stands in the start zone because the y-origin is below the start zone.

I hope this helps :P

### DIFF
--- a/scripting/include/shavit.inc
+++ b/scripting/include/shavit.inc
@@ -419,6 +419,15 @@ native int Shavit_ZoneExists(MapZones type);
 native int Shavit_InsideZone(int client, MapZones type);
 
 /**
+ * Checks if a player is inside a mapzone
+ *
+ * @param client					Client index.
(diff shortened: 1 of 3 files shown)


shavitush:
code doesn't compile, and i guess i'll just find my own solution soon

OfficialExtan:
oh okay ^^ worked for me but you can make a better solution :P I just copy pasted.

PR #46, opened 2016-06-08
Some recommendations

OfficialExtan:
-enable keys when client is on or in water

-enable keys when client uses noclip on a different style

-show client a message that he entered a glitchzone

What do you think?


OfficialExtan:
ignore the noclip commit xD was a mistake

PR #57, opened 2016-07-02
Implement rankings

shavitush:
+ Revamp some of `sm_stats`.
+ Add a whole new fair and competitive ranking system. Use `sm_setpoints <time in seconds> <points>`  to set points for an ideal time (usually \#1 or \#2 in the server).
+ Implement points in WR menus.
+ Implement points in mapsdone (`shavit-stats`).
+ Implement points in web interface.
+ Add some APIs, see `shavit.inc`

To-do before deleting branch:
- [ ] Add `shavit-chat` as said in README.md


Technoblazed:
YAY!

PR #69, opened 2016-07-12
Implement chat handler

shavitush:
Chat **(NEW!)**
--

Requires [Dynamic](https://forums.alliedmods.net/showthread.php?t=270519) to compile and run.

- [x] Add logic that processes chat without requiring an external plugin such as `Simple Chat Processor (Redux)`.
- [x] [RTLer](https://forums.alliedmods.net/showthread.php?p=1649882) support.
- [x] Custom chat titles/colors per individual player.
- [x] Custom chat titles/colors for rank ranges.
- [x] Update cache for a player when his rank updates.
- [x] Add `sm_ranks` `sm_chatranks`.
- [x] Add `Shavit_FormatChat` native.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,6 +117,7 @@ Stats
 Miscellaneous
 --
 - [x] Allow changing the prespeed limitation.
+- [x] Add weapon cleanup.
 
 Sounds **(NEW!)**
 --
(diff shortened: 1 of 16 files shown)


shavitush:
Opinions before I merge to master?

Technoblazed:
Taking a minute to look through the code, two secs :D

Technoblazed:
Everything seems good, how difficult would a % based ranking system be to implement into this?

shavitush:
Shouldn't be too hard (I think!), could you elaborate?

Technoblazed:
As it stands at the minute, this chat would do, i.e:

1 -> Master
2-5 -> Pro
7-15 -> Semipro Etc.

The percentage based system would be more along the lines of, 

1 -> Master
Top 1% -> Pro
Top 2-5% -> Semipri etc.

Where the % is the top of all players that use the server.

shavitush:
I see. I'll merge this for pull for now though. Would be cool if you started an issue asking for percentage-based rankings so it'll be easier to track. It will be probably a must for 1.5b.

PR #81, opened 2016-07-19
Add teleport zones

shavitush:
also tried to make them trigger based but i failed, left the code commented if anyone wants to try


shavitush:
Use sm_tpzone to define a teleport destination for this zone.

PR #89, opened 2016-07-24
New updates

shavitush:
Big stuff:
* Implemented Simple Chat Processor (Redux) support for shavit-chat, marked my chat processor as a fallback solution.
* Added the ability to use custom zone sprites and provided simple ones to begin with, setup configs/shavit-zones.cfg.
* Added `shavit_misc_noblock`
* Adjusted CS:S HUD to look just like bTimes (misses strafes/sync, will be added eventually) because apparently the community loves it.
* Replay bots will freeze on first/last frames, now they won't drop down randomly.
* Replay bots will die if they have no replay data loaded. Remove any auto-respawning cvar/plugin if you have one!
* HUD now shows the rank you're going to get on the map if you finish with your current time.
* Map rank added to the finish message.
* Added some natives/forwards, probably not useful for third-party modules.
* Added `sm_deletereplay` for RCON admins.
* Replay bots will automatically remove if a deletion of the \#1 record (style specific) happened.
* Hide Radar will also disable the radar in CS:S.
* `sm_rank` now also shows the total amount of ranked players. This is one step forwards on the road to percentage-based rankings.
* Timers won't start for non-prespeed styles if the player is jumping to them, this solves many issues like noclipping to a zone and keeping the speed. This will also make the timer start as soon as you jump (with blockprejump disabled) instead of when you leave the zone. Should've done before tbh..

# OPTIMIZED INSIDEZONE()!!!!!
Now your server won't shit itself as it did before with many players being alive. Did some profiling and there's a performance increase of at least 95% in a 1 slot server.


Technoblazed:
Hypu

PR #93, opened 2016-07-24
Binary Search Optimize

ofir753:
Binary Search Optimize


shavitush:
ended up not being as efficient as we thought

PR #208, opened 2016-09-10
Add styles configuration and deprecate shavit.inc configuration for styles

shavitush:
Will leave this pull request open for like 1-2 hours before I merge, looking for people to test. I tried myself and everything was backwards compatible and had 0 issues.

Edit styles in `addons/sourcemod/configs/shavit-styles.cfg`


Technoblazed:
I'll give it a go

Technoblazed:
I presume if you dont include the value, it defaults to default/0 ye?

Technoblazed:
Out of curiosity, what can using turn binds for be abused by in bhop to require a config option?

shavitush:
Default is 0 yeah, it says in the config file's header.  
Block moveleft is necessary for styles like W-Only because they only allow one key.  
I'll add a +left/right block setting too as 'legit servers' don't like them.

Technoblazed:
The header just says to start with 0, doesn't say what happens if you don't include the values in the style settings

shavitush:
```
// Start from 0 and increment every time. 0 is the default style.
```

Technoblazed:
Yeah, I got that, I think we're misunderstanding each other ;-;

Technoblazed:
I get 0 is the default, but in the actual styles itself, if you don't put a value in

shavitush:
It shouldn't cause a problem, but I'd really prefer that it would happen

shavitush:
Merged for now. Will also add some new features to this branch like the chat configuration file soon

Issue #223, opened 2016-09-11
Custom spawnpoint position


Nickelony:
e.g I want to spawn the player in the top left corner, not in the center of the zone. Could be a usefull feature, but you can do it after beta if you want.

Kryptanyte:
@shavitush I've already got this about half done right now, can make a pull request in an hour or two to add this feature.

shavitush:
@Kryptanyte sure! Make sure you add a cvar and follow the structure of the current code

Kryptanyte:
Having some trouble with my test server atm, gunna take a bit longer haha.

Nickelony:
When will this be finally added? :o

shavitush:
i kinda left a comment on the pull request, it's not left out :)

Kryptanyte:
Oh hey, sorry I had to deal with a drama of reinstalling windows, didn't actually see that comment until now o.o. I'll fix it up tomorrow. Sorry for the major delay

shavitush:
Solved withiin #261.

PR #237, opened 2016-09-12
Custom Spawn Location Addition

Kryptanyte:
**Custom Spawn Location**: #223 

Haven't super thoroughly tested this since I had a lot of problems with my test server but the only issue I could find is if you try to place a spawn when one already exists it errors, this error can crash the server after a few attempts.

I've worked around the bug by now allowing you to place a spawn if one already exists, you have to delete it first. I'll work on a fix for this tomorrow (It's 6 am right now).

I didn't add a cvar for it (yet) as restarting will default to the start zone which has a cvar for it as well. If you want me to add a check to see if the custom spawn is in the start zone or not I can, it's no hassle.

## Commands:

- sm_addspawn    **Adds a custom spawn location**
- sm_delspawn     **Deletes a custom spawn location**

### DIFF
--- a/scripting/shavit-zones.sp
+++ b/scripting/shavit-zones.sp
@@ -684,7 +684,7 @@ public Action Command_AddSpawn(int client, int args)
 
 	if(!IsPlayerAlive(client))
 	{
-		ReplyToCommand(client, "%s You can't place zones when you're dead.", PREFIX);
+		ReplyToCommand(client, "You can't place zones when you're dead.");
 
 		return Plugin_Handled;
 	}
--- a/scripting/shavit-zones.sp
+++ b/scripting/shavit-zones.sp
@@ -111,7 +111,6 @@ float gF_FreeStyleConstCos[MULTIPLEZONES_LIMIT];
 float gF_FreeStyleMinusConstCos[MULTIPLEZONES_LIMIT];
 
 bool gB_CustomSpawn;
-bool gB_UpdateCustomSpawn[MAXPLAYERS+1];
 float gF_CustomSpawn[3];
 
 float gF_RotateAngle[MAXPLAYERS+1];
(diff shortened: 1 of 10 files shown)


scripting/shavit-zones.sp
```diff
@@ -545,14 +554,23 @@ public void SQL_RefreshZones_Callback(Database db, DBResultSet results, const ch
...
 			{
 				continue;
 			}
 
+			if(type == Zone_CustomSpawn)
+			{
+				gF_CustomSpawn[0] = results.FetchFloat(1);
+				gF_CustomSpawn[1] = results.FetchFloat(2);
+				gF_CustomSpawn[2] = results.FetchFloat(3);
+				gB_CustomSpawn = true;
```
shavitush: Where is this being set to false?

scripting/shavit-zones.sp
```diff
@@ -642,6 +660,65 @@ public Action Command_TPZone(int client, int args)
...
+public Action Command_AddSpawn(int client, int args)
+{
+	if(!IsValidClient(client))
+	{
+		return Plugin_Handled;
+	}
+
+	if(!IsPlayerAlive(client))
+	{
+		ReplyToCommand(client, "%s You can't place zones when you're dead.", PREFIX);
```
shavitush: Please follow the new format of `Shavit_PrintToChat`

scripting/shavit-zones.sp
```diff
@@ -642,6 +660,65 @@ public Action Command_TPZone(int client, int args)
...
+
+		return Plugin_Handled;
+	}
+
+	gMZ_Type[client] = Zone_CustomSpawn;
+	GetClientAbsOrigin(client, gV_Point1[client]);
+
+	if(gB_CustomSpawn)
```
shavitush: No code block?

scripting/shavit-zones.sp
Kryptanyte:
```
public void ClearCustomSpawn()
{
	for(int i = 0; i < 3; i++)
	{
		gF_CustomSpawn[i] = 0.0;
	}
	gB_CustomSpawn = false;
}
```

Line 744 - 751. 
This function get's called in UnloadZones and the SQL_DeleteCustom_Spawn_Callback callback.
Thought it would be more efficient only having one for loop written instead of 3

scripting/shavit-zones.sp
Kryptanyte: Not sure what you mean? This is part of Command_Addspawn, also you're looking at an older commit?

scripting/shavit-zones.sp
Kryptanyte: Done

shavitush:
Could you

1. Update from the master branch.
2. Squash all the commits into one so I can review them.

Thanks for contributing :)

PR #336, opened 2017-03-31
fix replay bots getting stuck

PaxPlay:
Fixed Replay Bots getting stuck by teeleporting the bot every _ReplayCheckFrame_ frames.

### DIFF
--- a/scripting/shavit-replay.sp
+++ b/scripting/shavit-replay.sp
@@ -27,6 +27,8 @@
 #include <shavit>
 
 #define REPLAY_FORMAT_V2 "{SHAVITREPLAYFORMAT}{V2}"
+#define ReplayCheckFrame 100
+
 
 #pragma semicolon 1
 #pragma dynamic 131072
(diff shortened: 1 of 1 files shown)


PaxPlay:
Replay bots get laggy...

Zipcore:
if((gI_ReplayTick[iReplayBotStyle] % 128) == 0 && SquareRoot(Pow(vecVelocity[0], 2.0) + Pow(vecVelocity[1], 2.0)) < 2.0*Entity_GetDistanceOrigin(client, vecPosition))
		TeleportEntity(client, vecPosition, vecAngles, vecVelocity);
	else TeleportEntity(client, NULL_VECTOR, vecAngles, vecVelocity);

	stock float Entity_GetDistanceOrigin(int entity, const float vec[3])
	{
		float entityVec[3];
		GetEntPropVector(entity, Prop_Send, "m_vecOrigin", entityVec);
	
		return GetVectorDistance(entityVec, vec);
	}

Zipcore:
I've added support for high speed aswell, enjoy

shavitush:
might want to change 128 to `gF_Tickrate`

Zipcore:
Not really it doesn't have to be set to the tickrate. You could check everytime too so the bot can't get stuck for 1s or more.

Nickelony:
I would apply zipcore's fix tbh

shavitush:
I forgot about it! Will do today.

SlidyBat:
Does everything need to be square rooted? Bit expensive for OnPlayerRunCmd

shavitush:
My implementation doesn't do that. I use [GetVectorLength\(\)](https://github.com/alliedmodders/sourcemod/blob/237db0504c7a59e394828446af3e8ca3d53ef647/core/smn_vector.cpp#L45-L59) instead.

SlidyBat:
Both GetVectorLength and GetVectorDistance still square root by default, theres an optional parameter to leave them squared: https://sm.alliedmods.net/new-api/vector/GetVectorLength , https://sm.alliedmods.net/new-api/vector/GetVectorDistance

I think when just comparing, youre better off leaving squared

PR #368, opened 2017-06-20
Add shavit_misc_removeragdolls

strafe:
I've seen this suggested in earlier issues and declined as there are other plugins available to do this, however; for the timer to be as all-in-one as possible I think this is definitely a key feature to include.


shavitush:
Thanks for the contribution!

Issue #318, opened 2016-11-17
Translation issues + shavit_misc_noclipme "2" doesn't require the "ADMFLAG_CHEATS" flag


Nickelony:
V V V Scroll down V V V

Technoblazed:
YOU SAID THERE WERE NO MORE

DJSIFHBKADUSFGKUADADGKUFGADFHAFHH

Technoblazed:
I'll fix now

Technoblazed:
1) Check your translations:
```
FormatEx(sDisplay, 16, "%T", "GridSnap", client, gI_GridSnap[client]);
```
```
"GridSnap"
	{
		"#format"	"{1:d}"
		"en"		"Grid snap: x{1}"
	}
```

Technoblazed:
2) Check your translations:
```
"ZoneSizeIncrease"
	{
		"#format"	"{1:c},{2:c},{3:c},{4:d},{5:c},{6:.01f},{7:c}"
		"en"		"{1}{2} axis{3} (point {4}) increased by {5}{6}{7}"
	}
	"ZoneSizeDecrease"
	{
		"#format"	"{1:c},{2:c},{3:c},{4:d},{5:c},{6:.01f},{7:c}"
		"en"		"{1}{2} axis{3} (point {4}) decreased by {5}{6}{7}"
	}
```
Move your order around for value {2} and axis if you want to switch the values.

Nickelony:
1)
```
"GridSnap"
{
	"#format"	"{1:d}"
	"pl"		"Szerokość kratki: x{1}"
}
```

2)
```
"ZoneSizeIncrease"
{
	"#format"	"{1:c},{2:c},{3:c},{4:d},{5:c},{6:.01f},{7:c}"
	"pl"		"{1}{2} oś{3} (punkt {4}) została zwiększona o {5}{6}{7}"
}
"ZoneSizeDecrease"
{
	"#format"	"{1:c},{2:c},{3:c},{4:d},{5:c},{6:.01f},{7:c}"
	"pl"		"{1}{2} oś{3} (punkt {4}) została zmniejszona o {5}{6}{7}"
}
```

don't have much time now, sry :/

Technoblazed:
2)
```
"ZoneSizeIncrease"
{
    "#format"   "{1:c},{2:c},{3:c},{4:d},{5:c},{6:.01f},{7:c}"
[...]

PR #371, opened 2017-06-23
Fix CS:GO HUD indentation

strafe:
This was mentioned in #318, whether the timer is running or paused the player's best will be indented 1 tab too far, this seems to happen no matter the player's current running time/position.


Nickelony:
does it fix the problem? :o

strafe:
@Nickelony of course 😃

shavitush:
Thanks for contributing again! :)

PR #372, opened 2017-06-23
Zones plugin rewrite

### LINKED ISSUE #161: Rewrite most of the zones plugin
The current one is a mess, @ofir753 had a big part in it and it's pretty much a clusterfuck for me because I don't understand any of the basic math being done there and it makes it hard for me to even add new stuff because of it :/

So.. I need to rewrite nearly all of it.

Important stuff:
* ~~Get rid of the `MapZones` type and straight up use `int` instead. The current method is messy and I cast from `MapZones` to `int` (and vice versa) too often.~~
* ~~Make a hard limit of 64 zones.~~
* ~~Cache zones properly with Dynamic or `any` types for the sake of simplicity.~~
* ~~Use `trigger_multiple` for zones instead of the current implementation.~~
* ~~Fully optimize zone drawing, it's a big mess as of now.~~
* ~~Needs to be compatible with the current database structure or either have it auto-migrate to a new format.~~
* ~~Allow every zone to be created more than once, except for start/end (for main and bonus).~~

Note: This will get rid of zone rotations. The math is too heavy to run per-tick and per-client which kills server performance. This plugin was meant to be simple.. but the zones plugin doesn't allow it to be simple :(.

shavitush:
- Fixes #161.

I have restructured the whole zones plugin which will result in MASSIVE optimizations and ease of use, including:
* Tracks are now a thing. There isn't any implementation for them in other modules yet, but the backend is mostly ready.
* The whole code is much much better. It's easier to read, to understand and to work with.
* 3D boxes will not be generated often at all, but instead only once (when the zones are loaded from the database).
* The hard limit for zones right now is 64, but shouldn't be an issue to increase. AFAIK Source limits entities to a certain amount, and now zones are based on `trigger_multiple` so it's a failsafe. The value can be safely increased from within `shavit.inc` for CS:GO as the hard cap for entities is 4096 there.
* Zones use `trigger_multiple`s now, which results in lower CPU usage. Should make everything feel much more precise and responsive.
* Zones are now modular. There are forwards for entering and leaving zones too. See `Shavit_OnEnterZone` and `Shavit_OnLeaveZone`. New zones will be extremely easy to add.
* I have dropped support for drawing of flat zones. This is honestly because I'm absolutely crap at anything that has to do with basic math, and the previous solution would just flat out all the 8 zone points which is bad.
* There's an enumeration for zone cache, so I can add zone settings later on easily without any big refactors.
* You can now create multiple zones of each kind (except for start/end zone).
* I've dropped support for rotated/fixed zones due to working with `trigger_multiple` instead of pure math now.

Expect much higher server performance.

I'd love to see some testing before I merge this huge pull request! Please leave your feedback and report any issue/incompatibility that you may find.

### DIFF
--- a/configs/shavit-zones.cfg
+++ b/configs/shavit-zones.cfg
@@ -1,7 +1,7 @@
 // Absolutely DON'T delete subkeys from this file.
 //
 // visible defaults to 1
-// other values default to 255
+// other values default to 255, except for width which is 2.0
 "Zones"
 {
 	// zone sprites
@@ -26,6 +26,7 @@
 			"red"		"67"
 			"green"		"210"
 			"blue"		"230"
+			"width"		"2.0"
 		}
 
 		"End"
@@ -82,6 +83,7 @@
 			"visible"	"0"
 			"green"		"200"
 			"blue"		"0"
+			"width"		"4.0"
 		}
 	}
 }
--- a/scripting/include/shavit.inc
+++ b/scripting/include/shavit.inc
@@ -411,6 +411,30 @@ forward void Shavit_OnDatabaseLoaded(Database db);
 forward void Shavit_OnChatConfigLoaded();
 
 /**
+ * Called when a player enters a zone.
+ *
+ * @param client					Client index.
+ * @param type						Zone type.
+ * @param track						Zone track.
+ * @param id						Zone ID.
+ * @param entity					Zone entity iD.
+ * @noreturn
+ */
+forward void Shavit_OnEnterZone(int client, int type, int track, int id, int entity);
+
+/**
+ * Called when a player leaves a zone.
+ *
+ * @param client					Client index.
+ * @param type						Zone type.
+ * @param track						Zone track.
+ * @param id						Zone ID.
+ * @param entity					Zone entity iD.
+ * @noreturn
+ */
+forward void Shavit_OnLeaveZone(int client, int type, int track, int id, int entity);
+
+/**
  * Returns the game type the server is running.
  *
  * @return							Game type. (See "enum ServerGame")
--- a/scripting/include/shavit.inc
+++ b/scripting/include/shavit.inc
@@ -222,6 +222,13 @@ enum
 	ZONECACHE_SIZE
 };
 
+enum
+{
+	Track_Main,
+	Track_Bonus,
+	TRACKS_SIZE
+};
+
 // let's not throw errors k?
 stock bool IsValidClient(int client, bool bAlive = false) // when bAlive is false = technical checks, when it's true = gameplay checks
 {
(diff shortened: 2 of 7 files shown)


Nickelony:
So there is no "Rotate zone" option anymore? If so, then there are more leftovers in the pharses file such as "ZoneRotate"

shavitush:
No, the complex zone rotation is deprecated due to working with triggers now. I'll get rid of the zone rotation in the phrases!

strafe:
@shavitush can you implement some kind of wall snapping in the new zones plugin so that zones will ignore the grid on the axis they are snapped on but still follow it on the others, will help on getting zones to go into the corner without adjusting them manually after.

Also as I said in the issue I think you need to adjust the bottom and top bounds of the zone by 36 units rather than 16 since the model is taller than it is wider.

shavitush:
@strafe do you have any map where that's an issue? I'd love to test it in-game and find a solution.

I did testing in-game (both CS:S and CS:GO), -16.0 and +16.0 seemed to be just perfect for zones - as long as they're placed properly.

strafe:
It's not really an issue just a feature, you can test it out with blacky's timer if you have a copy easily accessible. It simply makes zoning quicker and easier by letting a zone point snap into a corner where two walls and the floor meet.

As for the zone contraction are you sure you tested exiting the top and bottom of the zone? If the timer is going to be compatible with fall-zones (a lot of fun bonuses start with fall-zones) or maps where the start zone floor is a booster (bhop_ny is a good example of this) I think these adjustments should be accurate. Since a CS:GO model is 72 units high and 32 units high surely your contraction is perfect for any side walls of the zones but not the top and bottom ones.

shavitush:
> Since a CS:GO model is 72 units high and 32 units high surely your contraction is perfect for any side walls of the zones but not the top and bottom ones.

This is a bit impossible(?) to do as all I do is set minimums/maximums for the zone's trigger, unfortunately.

strafe:
Hopefully so I such at maths and vectors so I have no clue either.

shavitush:
@strafe - is this what you meant by snapping to walls? _bhop_24_ is the only map I found which may have this kind of a start zone, so I used it.

ighsryq

strafe:
@shavitush yep looks really good, only thing I'd still like to see is something Mehis added where the visual zones/beams are retracted a bit so they aren't hidden inside the wall. The whole retraction thing is already pretty confusing but I still think it would be nice to shrink the visual box a bit too.

shavitush:
I honestly have no clue how to begin with that. However, apparently if the width of the beam is thin, it doesn't look like the beam overlaps with the zone. I'll add a `width` setting for zone types.

strafe:
Not sure where or how he does it in the code so I'll research it myself and if you feel like you can add it then I guess we can go from there.

Something else I think could be re-added quite easily is the 2D zone support, it seems Mehis does it just by [checking if the zone should be 2D that could be a convar instead or whatever and then only draws the extra beams if it is.](https://github.com/TotallyMehis/Influx-Timer/blob/master/addons/sourcemod/scripting/influx_zones_beams.sp#L629)

shavitush:
Top's width is 2.0 and bottom 15.5, for reference.

screenshot

@strafe I'll try to do that.

shavitush:
So, any objections before I merge this pull request?

Nickelony:
Can you add a cvar to set the default zone height when drawing? That would be useful.

shavitush:
Height is snapped to the grid, not to the ceiling.

Nickelony:
Perfect! I think this can be merged.

PR #400, opened 2017-07-17
Remove current style text from style menu

strafe:
This isn't really needed anymore since a player's current style appears in their HUD and is greyed out in the menu. I've also changed the layout a little simply so it matches the !wr menu appearance.


shavitush:
Thanks for contributing!

PR #403, opened 2017-07-17
Add "Minimum Velocity" or "X VelRun" style

Nickelony:
The most popular style on our servers. It's a must have for a true speedrun server experience. It's also the only major style that is missing in your timer. Hope you like it, but if not, then it's ok. :)

You can change the shavit-styles.cfg comment to whatever you think is right.


shavitush:
Definitely! Thanks for the contribution 😃 

Note: All plugins should be recompiled as the new variable is added in-between the other settings, otherwise you may experience weird issues.

PR #404, opened 2017-07-18
Add "bonus_velocity" to the style settings

Nickelony:
I've discovered this yesterday by accident, so I tested it a little and tbh, IT'S AMAZING! You can set it to 20 and have a slightly higher gain per jump which feels very natural. You can also set it to 200 for something more challenging. Hope you like it! :)

Edit: Just fixed the order in shavit-core. Everything should be fine now.
Let me know if I should update shavit-styles.cfg to the newest version aswell.


shavitush:
Seems good, thanks for contributing!

Issue #426, opened 2017-07-26
Dynamically apply zone HUD gradients


shavitush:
The current one is hard-coded and it's not very smooth unfortunately.

PR #431, opened 2017-07-27
Add dynamic HUD gradient (#426)

strafe:
My attempt at making the rainbow HUD _dynamic_.

This approach pretty much uses maximum brightness and saturation values and then loops through every `shavit_hud_gradientstepsize` hue values. Using `shavit_hud_gradientstepsize 1` makes it very slow so personally I feel `15` is a smooth but still noticeable speed.

Initially I did try to use HSV so I could just go through all the hue values in that spectrum but that turned out to be very complicated to convert to HEX or even RGB, this approach has no down sides though as far as I know and is much lighter code-wise.

Also to note, I was initially using a convar to modify a repeating timer to dictate the speed of the gradient but even when using that and the step size the timer speed was just... useless, even using the lowest values it's far too slow and using a step size makes it much easier to adjust the speed while still giving you a large spectrum of options.

Here's a video with the default step size, some of the steps are quite pronounced but using smaller values makes it harder to see that it's actually changing, it takes around 10 seconds to do a full loop with the default step size: https://streamable.com/45mup

### DIFF
--- a/scripting/shavit-hud.sp
+++ b/scripting/shavit-hud.sp
@@ -40,19 +40,10 @@ bool gB_Zones = false;
 bool gB_BhopStats = false;
 bool gB_Sounds = false;
 
-// zone colors
-char gS_StartColors[][] =
-{
-	"ff0000", "ff4000", "ff7f00", "ffbf00", "ffff00", "00ff00", "00ff80", "00ffff", "0080ff", "0000ff"
-};
-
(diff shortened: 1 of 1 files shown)


e-slut:
Looks great! The last HUD could give someone a seizure.

shavitush:
I actually haven't imagined someone else would do it instead, wow. Thanks!

PR #432, opened 2017-07-27
Bugfixes/additions

shavitush:
No comments for this one, I'll try to keep the commits easy to read instead.


shavitush:
The bfcc337, 217f1d5 and 1573385 commits should stop most (if not all) situations where the server spams errors about the plugin. (Native is not bound/plugin that owns this native is paused etc..)

If you're able to produce any of this error spam again after applying those, let me know ASAP.

PR #435, opened 2017-07-29
Multiple tracks (#133)

### LINKED ISSUE #133: add bonus timer
Checklist:

- [x] Database migration (add `track` integer to zones table).
- [x] Zone setup, toggle bonus/main track from zone setup menu.
- [x] Insert bonus zones to database.
- [x] Cache zone bonus status, load them from database.
- [x] Draw bonus zones too.
- [x] Bonus zone deletion.
- [x] Add natives where needed.
- [x] HUD integration for CS:S. I need someone else to do the CS:GO part, as I don't play CS:GO.
- [x] `sm_b`, etc
- [x] Database migration (add `track` integer to records table).
- [x] `sm_bwr`, etc
- [x] Separate the leaderboards arrays.
- [x] Insert bonus records to database, separated from main records,
- [x] Figure out how to implement replay bots.
- [x] Menus, record deletion etc.

shavitush:
\- Fixes #133.

This pull request introduces support for multiple tracks.
The plugin will have 2 built-in tracks. Main and bonus. Any records/zones set before this pull request will be retroactively set themselves to the main track.

To access the bonus timer, simply enter a bonus track start zone while your timer is stopped, or use `sm_b`.

Done:
* Updated most (if not all) of the natives and forwards that need a track.
* Core, misc and zones parts are pretty much finished. Might have forgotten some things, hopefully not.
* HUD for CS:S is finished. It's simply the track's name at the last row of the HUD. I'm not sure if there's a need for something similar in CS:GO. If you can implement it (and make it look decent) for CS:GO, hit me up with a screenshot and I'll tell you if a pull request is fine.
* The WR plugin was pretty complicated but I believe it's mostly okay now.
* Stats is a bit easier. I made it include more than just the main track and added "shavit_stats_mvprankones_maintrack" as a setting. If set to 1 (default), only the main track's records will be counted towards the MVP rank 1 records.
* Replays are applied retroactively. Non-main track replays will have a _X suffix. Bonus is _1 and so on. You can delete/watch replays through the menu. You can only watch bonus replays through the central bot. Bonus support for the legacy bot is not planned. Bonus replays will be *saved* even if the central bot is disabled. Added `{track}` as a variable in `shavit-replay.cfg` and added `Shavit_GetReplayBotTrack`.

Things irrelevant to tracks:
* Reworked `sv_airaccelerate` manipulation to be much smoother, and better for higher pings.
* Made the bounding box of the zones' triggers' height to be 100% accurate for CS:S. CS:GO is left as is, and is fine.
* Added a message whenever you change a style to something with different `sv_airaccelerate` values.
* Allowed per-style `sv_enablebunnyhopping` values. See https://github.com/shavitush/bhoptimer/pull/435/commits/9f220c045903bdd717c095c4e81af8d866695f52.
* Improvements to checkpoints. Now duck status and player stamina are saved too to make segmented times fair. Added targetname saving too, so you can't just skip whole maps like bhop_horseshit_5.
* Added `shavit_misc_resettargetname` (default 1) so dumb skips like this won't be a thing: https://youtu.be/FvGbGyzWhUY?t=440
* Made `sm_hide` mute footsteps (CS:S only) and gun sounds of others - the latter is untested because I couldn't find anyone to help me test, but it should work!
* Dropped the web interface. See this commit's messsage: https://github.com/shavitush/bhoptimer/pull/435/commits/ddf09c0901b82142c0b605fb5c324f701e66e7de
* Admins won't show up on spectator lists anymore. It only applies to players with the override "admin_speclisthide" or the KICK flag.


shavitush:
I think this is finished. Needs some testing (especially in CS:GO) before I merge though!

Nickelony:
Make "// rankings" and "// special flags" start with capital letters in shavit-styles.cfg like this: https://puu.sh/wXTIu.png cuz I think you forgot about them.

PR #460, opened 2017-08-04
Add "shavit_misc_noblood" for hiding blood effects.

Nickelony:
Hides all the blood effects including decals, splashes and other particles. Especially useful on hard maps where everyone is sitting at the spawn, hitting each other just to make the map "slightly" more red. :>
Not to mention that blood decals can sometimes make maps slightly harder if you have a lot of them in one place.

### DIFF
--- a/scripting/shavit-misc.sp
+++ b/scripting/shavit-misc.sp
@@ -113,6 +113,7 @@ ConVar gCV_HideRadar = null;
 ConVar gCV_TeleportCommands = null;
 ConVar gCV_NoWeaponDrops = null;
 ConVar gCV_NoBlock = null;
+ConVar gCV_NoBlood = null;
 ConVar gCV_AutoRespawn = null;
 ConVar gCV_CreateSpawnPoints = null;
 ConVar gCV_DisableRadio = null;
@@ -141,6 +142,7 @@ bool gB_HideRadar = true;
 bool gB_TeleportCommands = true;
 bool gB_NoWeaponDrops = true;
 bool gB_NoBlock = true;
+bool gB_NoBlood = false;
 float gF_AutoRespawn = 1.5;
 int gI_CreateSpawnPoints = 32;
 bool gB_DisableRadio = false;
@@ -256,6 +258,8 @@ public void OnPluginStart()
 	HookEvent("weapon_fire", Weapon_Fire);
 	AddCommandListener(Command_Drop, "drop");
 	AddTempEntHook("Shotgun Shot", Shotgun_Shot);
+	AddTempEntHook("EffectDispatch", EffectDispatch);
+	AddTempEntHook("World Decal", WorldDecal);
 
 	// phrases
 	LoadTranslations("common.phrases");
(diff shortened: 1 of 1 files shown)


scripting/shavit-misc.sp
```diff
@@ -2085,6 +2092,92 @@ public Action Shotgun_Shot(const char[] te_name, const int[] Players, int numCli
...
 
+public Action EffectDispatch(const char[] te_name, const Players[], int numClients, float delay)
+{
+	int iEffectIndex = TE_ReadNum("m_iEffectName");
+	int nHitBox = TE_ReadNum("m_nHitBox");
+	char sEffectName[64];
+
+	GetEffectName(iEffectIndex, sEffectName, 64);
+
+	if(gB_NoBlood)
```
shavitush: Could you move this `if()` statement to the beginning of `EffectDispatch` and invert it? So it looks like this:

```sp
if(!gB_NoBlood)
{
	return Plugin_Continue;
}
```

scripting/shavit-misc.sp
```diff
@@ -2085,6 +2092,92 @@ public Action Shotgun_Shot(const char[] te_name, const int[] Players, int numCli
 	return Plugin_Stop;
 }
 
+public Action EffectDispatch(const char[] te_name, const Players[], int numClients, float delay)
+{
+	int iEffectIndex = TE_ReadNum("m_iEffectName");
+	int nHitBox = TE_ReadNum("m_nHitBox");
+	char sEffectName[64];
```
shavitush: Could you make this initialize like other `char[]`s?

```sp
char[] sEffectName = new char[32];
```

I also prefer 32 because I don't think there's any longer effect name. - do the same for the Get...Names!

scripting/shavit-misc.sp
shavitush: Could you move this `if()` statement to the beginning of `EffectDispatch` and invert it? So it looks like this:

```sp
if(!gB_NoBlood)
{
	return Plugin_Continue;
}
```

Same for `WorldDecal`.

scripting/shavit-misc.sp
```diff
@@ -2085,6 +2092,92 @@ public Action Shotgun_Shot(const char[] te_name, const int[] Players, int numCli
...
+	}
+
+	return Plugin_Continue;
+}
+
+public int GetParticleEffectName(int index, char[] sEffectName, int maxlen)
+{
+	int table = INVALID_STRING_TABLE;
+	
+	if (table == INVALID_STRING_TABLE)
```
shavitush: Could you remove the space between the `if` and `(`? It doesn't match the current code style. Same for `GetEffectName` and `GetDecalfName`.

scripting/shavit-misc.sp
```diff
@@ -2085,6 +2092,92 @@ public Action Shotgun_Shot(const char[] te_name, const int[] Players, int numCli
...
+			return Plugin_Handled;
+		}
+	}
+
+	return Plugin_Continue;
+}
+
+public int GetParticleEffectName(int index, char[] sEffectName, int maxlen)
+{
+	int table = INVALID_STRING_TABLE;
```
shavitush: I think this is better as a `static int`, same for `GetEffectName` and `GetDecalfName`.

scripting/shavit-misc.sp
```diff
@@ -2085,6 +2092,92 @@ public Action Shotgun_Shot(const char[] te_name, const int[] Players, int numCli
...
+		if(StrContains(sDecalName, "decals/blood") == 0 && StrContains(sDecalName, "_subrect") != -1)
+		{
+			return Plugin_Handled;
+		}
+	}
+
+	return Plugin_Continue;
+}
+
+public int GetParticleEffectName(int index, char[] sEffectName, int maxlen)
```
shavitush: This needs to be a private function, therefore `public` needs to be gone. Same for the other Get...Names.

Nickelony:
I think this is ready to be merged. :)

shavitush:
No :( you didn't get me right!

Instead of this part:
```sp
public int GetParticleEffectName(int index, char[] sEffectName, int maxlen)
{
 	int table = INVALID_STRING_TABLE;
```

I wanted it to be like so:
```sp
int GetParticleEffectName(int index, char[] sEffectName, int maxlen)
{
	static int table = INVALID_STRING_TABLE;
```

shavitush:
Perfect, thanks~

PR #463, opened 2017-08-05
Redesign of shavit-timelimit

Nickelony:
I've slightly redesigned the shavit-timelimit plugin because it wasn't very useful. Now it's something you SHOULD have on your server.

**New Features:**
+ Automatically sets "mp_ignore_round_win_conditions" to 1.
+ "ForceMapEnd" to Force the map to end after the timelimit (with translations support!).
+ More CVars.

**New CVars:**
+ "shavit_timelimit_dynamic" to disable the dynamic timelimit if someone doesn't want it.
+ "shavit_timelimit_forcemapend" to disable "ForceMapEnd".

I'm waiting for feedback. :)

### DIFF
--- a/scripting/shavit-misc.sp
+++ b/scripting/shavit-misc.sp
@@ -2081,7 +2081,17 @@ public Action Shotgun_Shot(const char[] te_name, const int[] Players, int numCli
 	TE_WriteVector("m_vecOrigin", temp);
 	TE_WriteFloat("m_vecAngles[0]", TE_ReadFloat("m_vecAngles[0]"));
 	TE_WriteFloat("m_vecAngles[1]", TE_ReadFloat("m_vecAngles[1]"));
-	TE_WriteNum("m_iWeaponID", TE_ReadNum("m_iWeaponID"));
+	
+	if(gEV_Type == Engine_CSS)
+	{
+		TE_WriteNum("m_iWeaponID", TE_ReadNum("m_iWeaponID"));
+	}
+
+	else
+	{
+		TE_WriteNum("m_weapon", TE_ReadNum("m_weapon"));
+	}
+        
 	TE_WriteNum("m_iMode", TE_ReadNum("m_iMode"));
 	TE_WriteNum("m_iSeed", TE_ReadNum("m_iSeed"));
 	TE_WriteNum("m_iPlayer", (client - 1));


strafe:
Why does the `CheckRemainingTime` function try to get a client index from the passed data? You're passing `INVALID_HANDLE` to it so you should probably do.
```SourcePawn
Shavit_PrintToChatAll("%T", "MinutesLeft" LANG_SERVER, "60");
```

This `hTmp` thing seems to be recycled from Zipcore's [Force map end](https://forums.alliedmods.net/showthread.php?p=2245847) and is probably not the best way to do it anymore. Also since it's being checked every second it's kind of a waste to call all that `FindConVar` code every single time, SourceMod has a function `GetMapTimeLimit` which you could use in `OnMapStart` and cache.

Nickelony:
@strafe The "Force map end" plugin doesn't work anymore, so I changed some things to make it work again. I will try to implement the new method tomorrow. Thanks for the help! :)

strafe:
The OP version does not work but the version posted on page 2 still works perfectly for me, all Zipcore updated was the `CSRoundEndReason` to `CSRoundEnd_Draw` due to a game update 2 years ago. Either way I think this is a good addition for the timer, nice 😄

Nickelony:
I tried to do it the other method, but I failed and I really don't know how to make it work. I'm still playing with SP, trying to learn it, but I'm still too noobish for that. Any help will be appreciated.

strafe:
Influx already has [a good implementation](https://github.com/TotallyMehis/Influx-Timer/blob/master/addons/sourcemod/scripting/influx_fixtimelimit.sp#L38), maybe you can adapt that.

Nickelony:
@strafe It works! Thanks! :)

Nickelony:
This is ready to be reviewed.

scripting/shavit-timelimit.sp
```diff
@@ -244,6 +275,50 @@ void SetLimit(int time)
...
+public Action CheckRemainingTime(Handle timer)
+{
+	int timelimit;
+
+	if(!GetMapTimeLimit(timelimit) || timelimit == 0)
+	{
+		return Plugin_Continue;
+	}
+
+	if(timelimit > 0)
```
Technoblazed: Isn't this part sort of pointless? You can't have a negative timelimit, and you check if the timelimit is zero above?

Nickelony: My bad.

Nickelony:
Just fixed a small thing.
I'm not sure if you're going to like it, because this PR changes the whole purpose of this plugin.
It's okay if you don't like it. ;)

strafe:
Maybe you should name your commits a bit more descriptively... they're going to look awful in the commit history because they're so vague.

Issue #504, opened 2017-10-31
GetMapTier and GetMapTiers


PaxPlay:
Usage maybe in a custom mapchooser plugin to display the maps with tier when voting.
native int GetMapTier(char[] mapname);
native StringMap GetMapTiers(); // returns all map tiers

shavitush:
Will be definitely worked on, thanks.

However `Shavit_GetMapTier` will be a `bool`, as not every map has a tier defined to it.

PR #507, opened 2017-11-04
Additions and fixes

shavitush:
New natives! As requested in #504

```sp
/**
 * Gets the map tier for a specified map.
 * Use the map's display name.
 *
 * @param map						Map to get the tier of.
 * @return							Map tier. 0 if no results were found.
 */
native int Shavit_GetMapTier(const char[] map);

/**
 * Gets a StringMap that contains all the cached map tiers.
 *
 * @return							StringMap with {const char: map, int: tier} structure.
 */
native StringMap Shavit_GetMapTiers();

```

Usage example:

```sp
#include <sourcemod>
#include <shavit>

public void OnPluginStart()
{
	RegConsoleCmd("sm_xd", xd);
}

public Action xd(int client, int args)
{
	char map[] = "de_nuke";
	int tier = Shavit_GetMapTier(map);

	ReplyToCommand(client, "[1] %s: %d", map, tier);

	StringMap maps = Shavit_GetMapTiers();

	if(maps.GetValue(map, tier))
	{
		ReplyToCommand(client, "[2] %s: %d", map, tier);
	}

	else
	{
		ReplyToCommand(client, "[2] %s has no entry", map);
	}

	delete maps;

	return Plugin_Handled;
}
```

* Fixed tier updates for the cache upon change.
* Fixed issue with `sm_recalcall` giving wrong percentage after map changes.
* Allowed admins to see others in spectator list.
* Fixed an error in rank-for-time calculation.
* Fixed !b teleporting to wrong spawn point.
* Fixed unranked message not showing up when using !rank.
* Fixed issues with replay playback where the bot gets massive amounts of speeds, gets stuck and kills the server's performance.
* Fixed issue for triggers breaking because of shavit-misc. Set `shavit_misc_resettargetname` to 0 if you're not on a new installation.
* Extended checkpoints so they work on spectators - you can now borrow someone else's run and try it!
* Added `oldwr` as a parameter to `Shavit_OnWorldRecord`. Also added `Shavit_GetTimeForRank`.
* (CS:S) Fixed not being able to draw more than 5 zones per frame.
* Fixed an issue where remote MySQL databases would not properly assign in shavit-chat.
* Fixed zones not appearing after changing to a map with more zones.
* Fixed !ccname and !ccmsg with no arguments not showing current setting.
* Fixed an issue that caused zones to be drawn twice therefore appearing with an incorrect color.
* Added `Shavit_ReloadReplay` and `Shavit_ReloadReplays` natives.
* Added replay bot integration with [Bunnyhop Statistics](https://github.com/shavitush/bhopstats) to simplify the process of replay analyzing.
* Added per-track beam colors. Edit `configs/shavit-zones.cfg` with "Bonus Start"/"Bonus End" entries below "Easybhop"!
* Added KZ buttons support going by the KZTimer standard.

### DIFF
--- a/addons/sourcemod/configs/shavit-zones.cfg
+++ b/addons/sourcemod/configs/shavit-zones.cfg
@@ -125,5 +125,29 @@
 			"alpha"		"175"
 			"width"		"2.5"
 		}
+
+		"Bonus Start"
+		{
+			"visible"	"1"
+			
+			"red"		"255"
+			"green"		"255"
+			"blue"		"255"
+			
+			"alpha"		"255"
+			"width"		"0.1"
+		}
+
+		"Bonus End"
+		{
+			"visible"	"1"
+			
+			"red"		"123"
+			"green"		"20"
+			"blue"		"250"
+			
+			"alpha"		"255"
+			"width"		"0.1"
+		}
 	}
 }
--- a/addons/sourcemod/scripting/include/shavit.inc
+++ b/addons/sourcemod/scripting/include/shavit.inc
@@ -950,6 +950,23 @@ native ArrayList Shavit_GetReplayData(int client);
 native void Shavit_StopChatSound();
 
 /**
+ * Gets the map tier for a specified map.
+ * Use the map's display name.
+ *
+ * @param map						Map to get the tier of.
+ * @return							Map tier. 0 if no results were found.
+ */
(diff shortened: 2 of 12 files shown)


strafe:
I can't unsee that typo, `Reloads a specific replay into the repla bot cache.`

PR #576, opened 2018-01-13
Account for track when respawning on restart (#554)

### LINKED ISSUE #554: !b on spec teleports to main start

strafe:
Wasn't sure about the code-styling on this one. Resolves #554.

### DIFF
--- a/addons/sourcemod/scripting/shavit-misc.sp
+++ b/addons/sourcemod/scripting/shavit-misc.sp
@@ -1942,7 +1942,7 @@ public void Shavit_OnWorldRecord(int client, int style, float time, int jumps, i
 	}
 }
 
-public void Shavit_OnRestart(int client)
+public void Shavit_OnRestart(int client, int track)
 {
 	if(!gB_RespawnOnRestart)
 	{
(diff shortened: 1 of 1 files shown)


addons/sourcemod/scripting/shavit-misc.sp
```diff
@@ -1969,13 +1969,18 @@ public void Shavit_OnRestart(int client)
...
 	}
 }
 
-public Action Respawn(Handle Timer, any data)
+public Action Respawn(Handle Timer, DataPack pack)
 {
-	int client = GetClientFromSerial(data);
+	pack.Reset();
+	int client = GetClientFromSerial(pack.ReadCell());
+	int track = pack.ReadCell();
```
shavitush: needs a `delete pack;` here

shavitush: My fault, sorry! The only mismatch with code styling here is that everywhere else in the code, I initialize variable with their default values. For example, `Handle` (such as `DataPack`) would be initialized as `= null`, even if it changes nothing, I'd like to keep the code style consistent.

Went through it again and everything else seems completely fine!

strafe:
All done 🎨

Issue #440, opened 2017-07-30
Use GetGameFrameTime() for time calculation


shavitush:
Pending SourceMod 1.9's stable release. https://github.com/alliedmodders/sourcemod/commit/44c744b8fd4da788c02275599136ae3

Not high priority at all. But it'll solve issues with server lag affecting times and will allow me to easily implement timescaling for TAS.

PR #583, opened 2018-01-17
Some updates after a while with none

### LINKED ISSUE #570: Allow a config setting to store replay data in a specific folder/server with details
I run two servers linked to the same database. One whitelisted, another public. Linking replay data is annoying because if a bot is saved on one server, it's not possible to replay it on another.

Is there a possibility to set up a webserver and have the timer save replay data there? As long as both servers are linked to the one database, saving there wouldn't be an issue. This means that running Whitelist/VIP servers will be viable with the same database, meaning users can build around that.

### LINKED ISSUE #574: Make it so !spec makes you spec'ing replaybot

shavitush:
# * Removed support for SourceMod 1.8. Use 1.9 or above.
* Made it so that you can configure the replay data folder; fixes #570.
* Fixed a formatting issue with the top-left HUD. Players could previously cause formatting issues with control characters there.
* Fixed an issue where the total records count on other maps' WR menu is incorrect.
* Made the grid visualization less opaque than before.
* Added an extra native for the replay's API: `Shavit_GetReplayBotType`, also added the `ReplayBotType` type.
* Changed !spec to automatically forward you to the replay bot as long as arguments aren't supplied; fixes #574.
* Added `float oldtime` as a parameter to the `Shavit_OnFinish` forward.
* Reworked checkpoints. Now they're not actually 'capped' but rather soft-capped at 1,000 to avoid filling up the whole heap. You can change this manually by editing the `MAX_CPS` definition in `shavit-misc`, but you'll probably want to increase the `#pragma dynamic ...` directive too.
* Introduced replay bot format version `0x02`. It now records and plays back more data. Notably, movetype and entity flags.
* Added the `shavit_hud_ticksperupdate` cvar to allow server operators to change the amount of frequency of HUD updates.
* Tweaked the CS:GO HUD, see 9b5e4ad's message for information.
* Added `{tr}` as a parameter for `shavit_misc_clantag`. It's the first letter of the track you're on, and is only filled if your track isn't the default.
* Added the scroll pad to the CS:GO HUD's key display. Requires [Bunnyhop Statistics](https://github.com/shavitush/bhopstats).
* Fixed the issue where zone offsets were wrong! This isn't a thing anymore:
image
* Applied fixes for timer restarts with maps where the zones bundled into them.
* Made the CS:S HUD show your time even when you're paused.
* Made the resume command teleport you back to the position you were in while pausing. This fixes an exploit.
* Implemented #440. Please test this branch out and let me know if something is off, as this might be critical.
* Allowed pausing in practice mode.
* Fixed top-left HUD flickering.
* Revamped `block_pstrafe`. Option 1 will now just block inconsistencies with buttons. Option 2 will stop the timer. 1 stays the default.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,3 @@
-# i have no idea AT ALL on how to use travis,
-# stolen from splewis: https://raw.githubusercontent.com/splewis/csgo-pug-setup/master/.travis.yml
-# i'll fail anyways, right? :(
 sudo: false
 
 addons:
@@ -8,7 +5,7 @@ addons:
         - lib32stdc++6  # needed for spcomp
 
 env:
-    - SMVERSION=1.8
+    - SMVERSION=1.9
 
 before_script:
     # install smbuilder
@@ -26,7 +23,6 @@ before_script:
     - chmod +x spcomp
     - PATH+=":$PWD"
 
-    # - cp ../../../scripting/include/* include/ # i don't think this is necessary anymore
     - cd ../../..
 
 script:
--- a/README.md
+++ b/README.md
@@ -18,10 +18,10 @@ Including a records system, map zones (start/end marks etc), bonuses, HUD with u
 
 # Requirements:
 * Steam version of Counter-Strike: Source or Counter-Strike: Global Offensive.
-* [SourceMod 1.8 or above](http://www.sourcemod.net/downloads.php)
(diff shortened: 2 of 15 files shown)


shavitush:
Merging this now. I'm going to try rushing the development, and hopefully finish the chat-rankings implementation this week.

PR #603, opened 2018-03-15
Chat ranks!

shavitush:
screenie

Chat ranks have finally arrived. It's pretty self explanatory.
Check `addons/sourcemod/configs/shavit-chat.cfg` and update your translation files for this to work.

Other stuff:

* Fixed restart not resetting your pause status.
* Fixed replay abortion after the playback is over. It used to replay the same demo over and over.
* Added setting 4 to `shavit_misc_prespeed`. It allows prespeeding but only up to the value of `shavit_misc_prestrafelimit`.
* Fixed checkpoint saving when not spectating a valid player.
* Added the ability to save checkpoints from the replay bot.
* Removed trigger hooking from replay bots. This will only be annoying in *very specific maps*. SAME, SAME, INU INU INU (you can remove triggers like yonkoma's with [Stripper:Source](https://forums.alliedmods.net/showthread.php?t=39439))
* Fixed a few issues with replay playback when going in `trigger_teleport`.
* Fixed !r taking you nowhere when there's no zone in some cases.
* Reset targetname upon restarting with practice mode. This might not be the ideal solution, but I hope it is. You could previously exploit the previous mechanic (thanks KiD Fearless for reporting!).
* Removed ancient `shavit-core` cvars where appropriate style settings exist instead: `shavit_core_autobhop`, `shavit_core_blockleftright`, `shavit_core_nostaminareset`.
* Updated minimum wishspeeds for HSW. It should feel smoother in some cases now!
* Added `Shavit_LogMessage` native.
* Reverted footsteps muting. I just realized it doesn't really work in CS:GO, and it makes CS:S scroll styles unplayable.
* Added logging to `shavit-wr` and `shavit-zones`. All admin actions from these modules are logged.

### DIFF
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -23,7 +23,8 @@ It also contains support for built-in map timers (KZ) and the [Fly](https://gith
 The chat plugin manipulates chat messages sent by players.  
 It includes custom chat names, tags, colors and all can be defined by the players/admins.  
 Admins need the chat flag, or the "shavit_chat" override (good for a donator perk).  
-There's a user-friendly command named !cchelp so the users can easily understand what's going on.
+There's a user-friendly command named !cchelp so the users can easily understand what's going on.  
+In addition, it integrates with rankings and allows you to have titles for players according to their ranking, relative ranking or points in the server using !chatranks.
 
 #### shavit-hud
 The HUD plugin is `bhoptimer`'s OSD frontend.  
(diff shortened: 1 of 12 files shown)


cytrus224:
image

shavitush:
There we go, I think it's done. Will merge after a version bump commit today if no issues are found.
I'm really excited about 1.5b almost being finished, and am glad that this is [the last feature](https://github.com/shavitush/bhoptimer/milestones/Release) I wanted.

PR #626, opened 2018-04-28
Typos

strafe:
Just a few. The `perfect` param doesn't exist in the upstream version of [`bhopstats.inc`](https://github.com/shavitush/bhopstats/blob/master/scripting/include/bhopstats.inc#L28-L35).


shavitush:
yikes my english is terrible

PR #670, opened 2018-09-18
Don't force clients to download in-game sounds

strafe:


Nairdaa:
Edit https://github.com/shavitush/bhoptimer/blob/master/addons/sourcemod/configs/shavit-sounds.cfg file...

strafe:
> Edit https://github.com/shavitush/bhoptimer/blob/master/addons/sourcemod/configs/shavit-sounds.cfg file...

What?

I'm going to close this because as Slidy pointed out some people might not keep the sound file(s) on the server itself and only on fastdl. I'm guessing the only way to go about this (assuming there isn't a way to determine if a sound is custom or not programatically) is to add more logic to the sounds config and let users explicitly mark sounds as custom (or not custom with backwards compatibility in mind).

Nairdaa:
There are only custom sounds to download in the downloadables table. If people don't have the default sounds (wich is in their game package since the pipe update, instead of the default /sounds folder), connecting to a server won't force them to download them, it will instead print a console error while trying to play one.

Nairdaa:
Not to mention you missed the {} in the if statement.

sneak-it:
{} is not required.

Nairdaa:
Yes, I'm now completely aware of that now after Adam educated me in a discussion we've had on the official timer's Discord.

PR #688, opened 2018-10-18
Fix newline

strafe:


shavitush:
you need to merge master into this because theres a commit history rn

PR #749, opened 2019-03-09
Start zone HUD adjustments

### LINKED ISSUE #748: [Suggestion] Change bonus hud start zone
Instead of:
Tier
In Start Zone
Speed

Should be:
BONUS
In Start Zone
Speed

Since the tier doesn't matter in the bonus start zone, there is no point in showing it and could be replaced "BONUS" or something.

strafe:
This replaces the tier text in non-main start zones with the track name, as well as indenting the first line to center it.

Preview:
image
image

Closes #748


shavitush:
big fan

PR #779, opened 2019-04-07
Update shavit-core.sp

Nairdaa:
Added !endb as many people asked me to do so on my server, so it's not only !bend.

### DIFF
--- a/addons/sourcemod/scripting/shavit-core.sp
+++ b/addons/sourcemod/scripting/shavit-core.sp
@@ -277,6 +277,7 @@ public void OnPluginStart()
 	RegConsoleCmd("sm_end", Command_TeleportEnd, "Teleport to endzone.");
 
 	RegConsoleCmd("sm_bend", Command_TeleportEnd, "Teleport to endzone of the bonus track.");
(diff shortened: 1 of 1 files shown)


addons/sourcemod/scripting/shavit-core.sp
```diff
@@ -277,6 +277,7 @@ public void OnPluginStart()
 	RegConsoleCmd("sm_end", Command_TeleportEnd, "Teleport to endzone.");
 
 	RegConsoleCmd("sm_bend", Command_TeleportEnd, "Teleport to endzone of the bonus track.");
+	RegConsoleCmd("sm_endb", Command_TeleportEnd, "Teleport to endzone of the bonus track.");
```
strafe: This won't work. All end commands (main and bonus) use the same callback, which checks for the existence of `sm_b` in the issued command to determine which track to use.

https://github.com/shavitush/bhoptimer/blob/f140bbb0ab0ba65d6a5d9085fcaec30ff5d01ef2/addons/sourcemod/scripting/shavit-core.sp#L461-L466

Nairdaa: Oh god damnit. My bad. :/

PR #802, opened 2019-05-23
Update shavit-core.sp

Nairdaa:
angle thing edit, preparing for HSW sync fix as well so it actually fkin works


Nairdaa:
Useless sanity check.

PR #808, opened 2019-06-02
Very good yes

Nairdaa:
added HSW support
should match shavit standards


Nairdaa:
Omg. Damn W+A doesn't work properly. W+D does. I'll do more testing -.-

PR #809, opened 2019-06-02
HSW Sync now works as it should.

Nairdaa:
Tested, fixed and working as intended. Enjoy.


shavitush:
used your implementation in dfcaee5, just wanted to split it to a different function! thanks a bunch

PR #867, opened 2020-03-25
Fix a bug

deadw1nter:
You still can save a segmetned checkpoint when you are dead via menu.
[how to reproduce](https://youtu.be/piPZ07-a3-o)


FlayX096:
ur in practice mode if u teleport to the cp so its just show just the time to you

PR #881, opened 2020-04-02
big bug

Nairdaa:
check if it works before pulling :(

// thanks sneakz for noticing i requested the wrong branch


kidfearless:
Game breaking issue right here :P

PR #886, opened 2020-04-09
Clean up prerun code + Add prerun countdown to hud

deadw1nter:


addons/sourcemod/scripting/shavit-replay.sp
```diff
@@ -701,6 +700,13 @@ public int Native_GetReplayTime(Handle handler, int numParams)
 		return view_as<int>(GetReplayLength(Track_Main, track));
 	}
 
+	if(FindConVar("shavit_hud_prerun_countdown").BoolValue)
```
kidfearless, line 703: The convar should be cached in the plugin

Issue #818, opened 2019-07-13
[Request] Dynamic Time Difference


blankbhop:
Fuckithops has a feature where it displays a time difference between the player and the current wr, I'm not sure how exactly it works, but I think it is something with the bot path and how close you are to the bot at a certain time.

Nairdaa:
Has nothing to do with the bot.
WR time - Current time = time to WR.

kidfearless:
added

Nairdaa:
heH

PR #889, opened 2020-04-14
Remove extra prerun frame, Add dynamic time difference

deadw1nter:
# Changes
- Fix an issue that prerun frame size is bigger than intended size when ``shavit_replay_prerun_always`` is 0
- Add dynamic time difference (#818)

# Something I can't do
- [ ] Polish native name and description


deadw1nter:
Closed because of wrong branch.

PR #918, opened 2020-05-31
Fix hud error bug spam in server console and prespeed on replay

Nairdaa:
Fully tested, no errors in compiling or in game or on the server. Works with TAS and any other segmented style as well.

Also, removed dynamic time difference. Shit was buggy. Going to be reworked and done way better.

Thanks alot, Deadwinter. Updating on behalf of him.


Nairdaa:
Oh, added the knife sound removal from !hide.

Nairdaa:
No errors so far.

PR #929, opened 2020-07-03
imagine hitboxes

Nairdaa:
rumour is a beast by showing the hitboxes and mev is even beasterest for catching this fuck up

### DIFF
--- a/addons/sourcemod/scripting/shavit-zones.sp
+++ b/addons/sourcemod/scripting/shavit-zones.sp
@@ -2887,14 +2887,14 @@ public void CreateZoneEntities()
 		float height = ((IsSource2013(gEV_Type))? 62.0:72.0) / 2;
 
 		float min[3];
-		min[0] = -distance_x + 16.0;
-		min[1] = -distance_y + 16.0;
+		min[0] = -distance_x;
+		min[1] = -distance_y;
 		min[2] = -distance_z + height;
 		SetEntPropVector(entity, Prop_Send, "m_vecMins", min);
 
 		float max[3];
-		max[0] = distance_x - 16.0;
-		max[1] = distance_y - 16.0;
+		max[0] = distance_x;
+		max[1] = distance_y;
 		max[2] = distance_z - height;
 		SetEntPropVector(entity, Prop_Send, "m_vecMaxs", max);
 


kidfearless:
What's this for

Nairdaa:
https://i.imgur.com/BUuo5Il.jpg
very bad



https://i.imgur.com/a2quJSV.png
very good


start the timer when your hitbox has actually left the zone, same shit applies for endzones

OSCAR-WOS:
this gathered so much stipulation due to users requesting the timer start when your eye (camera) leaves the start zone, rather than your body. this change would invalidate all records also.

Nairdaa:
> 
> 
> this gathered so much stipulation due to users requesting the timer start when your eye (camera) leaves the start zone, rather than your body. this change would invalidate all records also.

Yea better rely on something people established 10 years ago when they had no idea how the shit works and let's keep on having the bhoptimers counting the actual time wrong rather than having it done right. 

Following that logic, every timer with checkpoints should also stop counting the last checkpoint and start counting next at the same time the player's hitbox in the middle of the checkpoint line.

Nairdaa:
Allowing people to drop all their weapons for 260 walk speed also invalidated all the strafe map records.

OSCAR-WOS:
I hope you never receive modify access to this project

Nairdaa:
Quite unfortunate for you I already have a green light from this timer's dev to implement this as an optional config and I'll do it wether you like it or not. That's the way the timers should be counted, right when you leave the startzone and right when you touch the endzone.

Let me quote rumour on this one:
[https://www.justhorseracing.com.au/wp-content/uploads/bendigophoto-734x460.jpg](url)


also

> I hope you never receive modify access to this project

lmao, you are so mad

kidfearless:
> 
> 
> I hope you never receive modify access to this project

Can confirm Nairda's not getting modify access to the project XD.

In all honesty, I don't see many people implementing it on their servers, and most haven't. It's already been in the timer for years now in the form of `shavit_zones_offset`. If they really wanted zones to function that way they could always have modified that convar to match how they feel the zones should be.

The implementation as a convar is for people like Nairda who only thought about it the past week and found that they don't like it and want to change all their zones to match that method without querying their database.

As a side note, the defaults will be the same as before. I don't think many people will enjoy getting plummeted to the ground because they thought they were out of the start zone but their butts were still inside and therefore haven't fully left the start zone.

PR #935, opened 2020-07-10
Fix offsets, add offset natives for styles like TAS

hermansimensen:
Also doesn't calculate offsets when NoZAxisSpeed is set to 0 and player is jumping from within the startzone, as requested by rumour.

### DIFF
--- a/addons/sourcemod/scripting/shavit-core.sp
+++ b/addons/sourcemod/scripting/shavit-core.sp
@@ -33,7 +33,7 @@
 #pragma newdecls required
 #pragma semicolon 1
 
-// #define DEBUG
+// #define DEBUG 1
 
 enum struct playertimer_t
 {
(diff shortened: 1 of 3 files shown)


addons/sourcemod/scripting/shavit-core.sp
```diff
@@ -1621,6 +1631,31 @@ public int Native_PauseTimer(Handle handler, int numParams)
 	PauseTimer(client);
 }
 
+public int Native_GetTimeOffset(Handle handler, int numParams)
+{
+	int client = GetNativeCell(1);
+	int zonetype = GetNativeCell(2);
+	
+	if(zonetype < 1)
```
kidfearless, line 1639: can't really think of how you would write this a different way.
```if(!(zonetype == Zone_Start || zonetype == Zone_End))
{
    throw native error 200 index out of bounds
}
```

addons/sourcemod/scripting/shavit-core.sp
```diff
@@ -1621,6 +1631,31 @@ public int Native_PauseTimer(Handle handler, int numParams)
 	PauseTimer(client);
 }
 
+public int Native_GetTimeOffset(Handle handler, int numParams)
```
kidfearless, line 1634: make the native an any and return the float without the view as

PR #948, opened 2020-07-30
Fix offset bug and add GetStageZone native

hermansimensen:
more spaghetti


kidfearless:
blindly accepting spaghetti

Issue #793, opened 2019-04-28
Persist CP data


blankbhop:
After disconnection CP's should persist, along with Segmented run data. This is mandatory as I was in a sub 4 minute segmented run on bhop_nextgen, and I crashed and my entire run was gone.

Nairdaa:
> =＾● ⋏ ●＾=02/04/2019
> it's much more complicated to save cps
> i can try in next update, no guarantees

:-(

blankbhop:
This timer is a -rep until this is a feature :(

Nairdaa:
You have it wrong. It's a +rep, because that might become a feature. It's not like you even had the reconnect respawn feature on blackys anyway.

whocodes:
don’t use cheat and your game won’t crash HA

blankbhop:
Dude delete this comment, I can't bhop properly without the cheats

bhopalex:
This is a feature I'd like to suggest adding as well. Preferably with the option to enable it for players by flag.

PR #952, opened 2020-08-08
add persist checkpoint data

Nairdaa:
As always and forever, deadwinter is a pogchamp.

https://github.com/shavitush/bhoptimer/issues/793

### DIFF
--- a/addons/sourcemod/scripting/include/shavit.inc
+++ b/addons/sourcemod/scripting/include/shavit.inc
@@ -844,7 +844,7 @@ forward void Shavit_OnTimescaleChanged(int client, float oldtimescale, float new
 forward Action Shavit_OnCheckPointMenuMade(int client, bool segmented);
 
 /**
- * Called before a selection is processed in the main checkpoint menu.
+ * Called when before a selection is processed in the main checkpoint menu.
  *
  * @param client					Client index.
  * @param param2					Second parameter in the callback, usually the item selected.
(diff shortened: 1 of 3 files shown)


Nairdaa:
TODO:

shavit_misc_persistcpdata 1/0 
Soon™

kidfearless:
It looks like you started from an older commit and didn't update. As well as changed 2 other sections that aren't related to the intent of the PR. As it is now I'm reworking the PersistData section of the timer as it turns out cloning big arraylists causes some server lag. And adding checkpoint cloning on top of that definitely won't help.
```
Shavit_GetReplayData 82.362083(0.001060/0.001287)
props 1.010100(0.000012/0.001287)
snapshot 0.155400(0.000001/0.001287)
ClassNames 0.233100(0.000003/0.001287)
push 0.077700(0.000000/0.001287)
```

Nairdaa:
"This branch has no conflicts with the base branch"

deadw1nter:
> And adding checkpoint cloning on top of that definitely won't help.
If I don't clone the checkpoint array, it will not pass the array to aData.aCheckpoints. I don't really know how to avoid cloning array.

kidfearless:
you shouldn't need to clone it [here](https://github.com/shavitush/bhoptimer/pull/952/files#diff-1dd7198ac08ba95c7cc87cf4c8816466R1292). At that point the arraylist has already been cloned once and is being cloned again. You should just simply assign it to the variable.

I can't accept this pr for now, as it's going to more than likely end up conflicting with the changes I'm currently working on. 

For it to be accepted in the future though, you'll need to set up shavit-misc to not delete the checkpoint(s) that will persist when the client leaves. Instead add them to the list of persistent checkpoints and null the players checkpoint variables. This will save their checkpoint(s) without having to clone them. 
On top of that you'll need to remove those changes that aren't part of the pr.

PR #963, opened 2020-10-15
Allow !goto/!tpto from start/end zone

rtldg:
The player's timer is stopped before teleporting the client too so dw about anything


kidfearless:
This has the potential to cause issues and will need to be looked into more before merging.

Nairdaa:
This is not allowed on purpose, because it can be abused pretty easily. Won't say in public how tho.

rtldg:
@Nairdaa That sounds like it's already a problem then.

Also the Shavit_StopTimer/Native_StopTimer function stops the timer since it's not blockable via the Shavit_OnStopPre forward as bBypass is true with numParams < 2

PR #964, opened 2020-10-15
[CSS] continuously set m_flFlashDuration so the radar goes away

rtldg:
With the radar disabled, it'll come back after an hour or if you lag.
You can replicate this by... waiting an hour, or with
```
sm_rcon sv_cheats 1
net_droppackets 200
```
The `GetURandomFloat()` call is used because the radar won't go away if you set the m_flFlashDuration to the same thing over and over.

### DIFF
--- a/addons/sourcemod/scripting/shavit-misc.sp
+++ b/addons/sourcemod/scripting/shavit-misc.sp
@@ -763,9 +763,15 @@ public MRESReturn CCSPlayer__GetPlayerMaxSpeed(int pThis, Handle hReturn)
 public Action Timer_Cron(Handle Timer)
 {
 	if(gCV_HideRadar.BoolValue)
-		for (int i = 1; i <= MaxClients; i++)
-			if (IsValidClient(i))
+	{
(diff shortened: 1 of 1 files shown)


rtldg:
sorry for the conflicts, I kind of expected that to happen since both this and the persistdata pr both put stuff at the top of Timer_Cron

Nairdaa:
@rtldg While you are here, fixing bugs, could you perhaps take a look at the cp system? Sometimes when you save a checkpoint and teleport to it, it makes you unavailable to land on the next block, the trigger resets you just like if you stood still on that block.

There is 
>SetEntPropString(client, Prop_Data, "m_iName", sTargetname);

Yet, it bugs out and no1 knows why apparently.

kidfearless:
> 
> 
> sorry for the conflicts, I kind of expected that to happen since both this and the persistdata pr both put stuff at the top of Timer_Cron

It's fine, they were all broken up into smaller issues making it easier to manage. Thanks for fixing all those bugs.

PR #965, opened 2020-10-15
only print the stage time message once per stage

rtldg:


kidfearless:
This was intentional so that players running segmented could see their time difference while grinding. With the current change a checkpointing player would need to restart his run every time he goes through the stage zone.

rtldg:
What about making it print every time for segmented then? (And TAS whenever hypnos decides that is ready to merge)

Nairdaa:
> What about making it print every time for segmented then? (And TAS whenever hypnos decides that is ready to merge)

That would be the best idea.

kidfearless:
The best thing to do would be to save their last stage for every checkpoint saved, and then reset it to the saved value when they teleport.

Nairdaa:
> The best thing to do would be to save their last stage for every checkpoint saved, and then reset it to the saved value when they teleport.

No need to store even more data in the array. What rtldg suggested is just an if statement and to me seems to be the best solution.

PR #966, opened 2020-10-15
allow !resume when not on the ground

rtldg:
this was done because someone paused in the tunnel on badges, went into spec, and came back in the void causing them to not be able to resume

unsure if this could be abused (any more than !pause could be already)
perhaps pausing as you slide off a platform, into a trigger, resuming and then getting a boost?
are any maps like that?


Nairdaa:
> are any maps like that?

Many. I wouldn't implement this change.

kidfearless:
This has the potential to cause issues and will need to be looked into more before merging.

Issue #953, opened 2020-08-10
Dynamic time diff bug


Nairdaa:
A reminder for our current lazy dev. He knows what to do, but is too busy at the moment.

Steps taken to reproduce: Play, watch the timer suddenly going from for example -0.1 sec to +20 sec and stay +x forever (counting, but wrong).

Replay bots:
bhop_mae
Positon: setpos 1266.043823 423.967499 -29.968689;setang 19.866026 171.646530 0.000000
Style: Scroll

bhop_badges
Starts acting already outside of the startzone
Style: Normal

[bhop_mae.zip](https://github.com/shavitush/bhoptimer/files/5048771/bhop_mae.zip)
[bhop_badges.zip](https://github.com/shavitush/bhoptimer/files/5048774/bhop_badges.zip)

Let me know if you need more. That kind of stuff never happened on log's approach to dynamic time difference.

Nairdaa:
[minusnotrng_plusrng.zip](https://github.com/shavitush/bhoptimer/files/5048793/minusnotrng_plusrng.zip) example demo.

kidfearless:
I'm unable to reproduce it and have a nagging feeling that you are still running the version that was broken.
https://youtu.be/7pxriZA4jFc
The replay was 2665 frames long at 100 tick giving it a 26 second time. I finished with a 22 second time and a dynamic time difference of around 4 seconds.
Tested on the latest [very_good_yes](https://github.com/shavitush/bhoptimer/tree/08050e23acb81f8a12a5a67d1e972d45134b9bda)

Nairdaa:
Downloaded latest, compiled, tried again it's still fucked up. Told you to come and see how to reproduce, but you still don't know how to reproduce that even tho I recorded a video for you. Come and see it live how to cause it.

Nairdaa:
[ Heroine ] kebenjekajse : now
♫♫ an idiot sandwich : what happened
[ Heroine ] kebenjekajse : said -10s
[ Heroine ] kebenjekajse : when i started 
♫♫ an idiot sandwich : k
[ Heroine ] kebenjekajse : it went away now
♫♫ an idiot sandwich : disappeared?
[ Heroine ] kebenjekajse : yea but back again now :D

Nairdaa:
Also, after updating it's not as fucky as it used to, but still fucks up sometimes.

Nairdaa:
[...]

PR #970, opened 2020-10-17
grab Z from replay origin for dynamic time diff

rtldg:
This is a part of the fix for #953


kidfearless:
Should probably be implemented as a client preference like 2d Vs 3d velocity in the HUD. Either that or a convar.

rtldg:
I don't think that should be an option as it'd just be an option for having a broken time difference.

kidfearless:
It would be for reducing the variation from the replay being at the peak of their jump and the player being at the start of their jump.

PR #971, opened 2020-10-17
Don't include preframe count for dynamic time diff

rtldg:
https://github.com/shavitush/bhoptimer/issues/953#issuecomment-674516766
"Also, when you go back to startzone (not with !r) your dynamic time is not back to 0 when you have more than 10 seconds on the timer (you can have even 30 on it, walk back to startzone, walk out and it will show -10 sec on dynamic timer right away)."

Someone else test this patch too @Nairdaa


Nairdaa:
It's certainly better but absolutely doesn't fix the issue and how the sorting is done. The way it was done previously by Log was much more accurate and cpu friendly. You could just stand there for 2 hours and the timer would show the difference properly, why this literally rapes a dedic with "shavit_replay_dynamictimedifference_search "10.0"" changed to even 1600 as it's hardcoded in the timer.

Issue #770, opened 2019-03-30
[ Suggestion ] !wr


Nairdaa:
Hello, 

Atm when you type whatever you want after "!wr" a record menu will display with the string given after "!wr" anyway regardless of whether the map exists on the server or not. 

Could you please make a check and print in the chat "Map not found" just the way it works with !nominate %s ?

Nairdaa:
Also there is another thing with !wr

When you type !wr it shows the wr time.
When you type !wr %s (some other map) it won't show the wr times on it.

Nairdaa:
https://github.com/shavitush/bhoptimer/pull/972

Soon™

PR #972, opened 2020-10-17
Wr suggestions

rtldg:
Made `!wr %s` & `!tier %s` with a non-existent map print `Map %s was not found` from #770
Also deduplicated some functions.


kidfearless:
generally I prefer not to overload functions by doing string comparisons since I had performance issues with bash2 due to chained string comparisons.

That said it's 4 characters and will be fine, good work 👍

kidfearless:
Oh also, the formatting style has it so that unless you're modifying a variable, you don't pass by reference.

rtldg:
I have no idea if an ArrayList is copied when used as a parameter in sourcepawn, but since it compiled when using a reference it was fine not having to worry :^)

Issue #944, opened 2020-07-27
Block SHSW on HSW style.


cytrus224:
blame nairda for this being not fixed yet

PR #973, opened 2020-10-17
block s on regular hsw (so you can't shsw...)

rtldg:
~~Force block `S`~~ #944
Block SHSW on HSW while still allowing S to be used in the air


Nairdaa:
I disagree with this change. You are supposed to be able to stop on certain maps (eg. badges spike). It's better to do it old fashioned way to just allow pressing S like once every "x" strafes in my opinion.

Nairdaa:
Basically, alow to S when W/A/D are not being pressed and release S key if any of the other keys are pressed too.

Nairdaa:
Working completely fine, the change has been on Ultima WL for over 2 months.

Issue #816, opened 2019-07-06
[Request] Add turn binds to sm_showkeys


mbhound:
+left
+right

Nairdaa:
Could you please add side movement direction with arrows to !keys too? Like it's on btimes, moving mouse to the left showing left arrow and accordingly to moving right.

blankbhop:
bTimes also has a weird issue with it showing mouse movement whenever somebody is spectating, be weary of that if implementing.

rtldg:
> bTimes also has a weird issue with it showing mouse movement whenever somebody is spectating, be weary of that if implementing.

I think I see what you're talking about.
https://github.com/blankbhop/btimes/blob/e8049b01b7e3eb1751e3f6c6de49a9e71d2c1cf4/addons/sourcemod/scripting/bTimes-hud-css.sp#L611
Sets the g_fOldAngle[target] and then any keys display after wouldn't work.

blankbhop:
The key display itself works as intended, the actual arrow representing the angle difference is what looks wrong whenever somebody is spectating the player.

rtldg:
> Could you please add side movement direction with arrows to !keys too? Like it's on btimes, moving mouse to the left showing left arrow and accordingly to moving right.

done't in https://github.com/shavitush/bhoptimer/commit/3750c8edebb2d7a590b75dff9e64af836a592792 although idk if it looks good yet because the characters don't render for me xd

PR #980, opened 2020-10-28
add turn binds to !keys

rtldg:
#816


kidfearless:
Can you post an image of what it looks like?

PR #988, opened 2020-11-30
Revert "Allow !goto/!tpto from start/end zone"

Nairdaa:
Reverts shavitush/bhoptimer#963

It can get abused easily. Don't.


Nairdaa:
Not anymore. The option to abuse it has been fixed by acer. Idk if it's pr'ed yet or not.

PR #989, opened 2020-11-30
Added SetStart command to allow players to change their spawn location in the start zone

BoomShotKapow:
I apologize if this isn't the proper way of contributing to the timer, but I mostly just wanted to familiarize myself with SourcePawn and this was a feature I wanted to see in the timer.

### DIFF
--- a/addons/sourcemod/scripting/shavit-core.sp
+++ b/addons/sourcemod/scripting/shavit-core.sp
@@ -77,7 +77,6 @@ bool gB_MySQL = false;
 
 // forwards
 Handle gH_Forwards_Start = null;
-Handle gH_Forwards_SetStart = null;
 Handle gH_Forwards_Stop = null;
 Handle gH_Forwards_StopPre = null;
 Handle gH_Forwards_FinishPre = null;
@@ -229,7 +228,6 @@ public APLRes AskPluginLoad2(Handle myself, bool late, char[] error, int err_max
 	CreateNative("Shavit_SaveSnapshot", Native_SaveSnapshot);
 	CreateNative("Shavit_SetPracticeMode", Native_SetPracticeMode);
 	CreateNative("Shavit_StartTimer", Native_StartTimer);
-	CreateNative("Shavit_SetStart", Native_SetStart);
 	CreateNative("Shavit_StopChatSound", Native_StopChatSound);
 	CreateNative("Shavit_StopTimer", Native_StopTimer);
 	CreateNative("Shavit_GetClientTimescale", Native_GetClientTimescale);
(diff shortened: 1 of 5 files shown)


Nairdaa:
https://gist.github.com/Nickelony/b62e01f41b4c39d60ce7ba80f43cb79c


It's already somewhere in the resources on discord.

Nairdaa:
https://discord.com/channels/389675819959844865/495188255604473856/722168119249993908

There it is. 

https://discord.com/channels/389675819959844865/495188255604473856/722168119249993908

BoomShotKapow:
Oh well, it was fun to do and familiarize myself with SourcePawn, since I've never used it before. Thanks for letting me know about those plugins though,

Nairdaa:
I mean don't get me wrong. You did a good job, but I don't think the devs want it merged. I'm for if you ask me.

BoomShotKapow:
Yea, I wasn't sure how to contribute. It's why I did a draft pull request, but if any of it is useful, it's there for them.

addons/sourcemod/scripting/shavit-core.sp
```diff
@@ -2068,6 +2113,19 @@ void StartTimer(int client, int track)
...
 }
 
+void SetStart(int client)
+{
+	if(!IsValidClient(client) || IsFakeClient(client))
+	{
+		return;
+	}
+	
+	Call_StartForward(gH_Forwards_SetStart);
```
kidfearless, line 2123: Why did you implement this as a forward? It'd make more sense to implement a native in shavit-zones that would handle that.

kidfearless (changes requested):
> Yea, I wasn't sure how to contribute. It's why I did a draft pull request, but if any of it is useful, it's there for them.
contributions are always appreciated. A lot of people have used the set start plugin from the discord, and the changes to save start positions were also talked about a while back, but I didn't feel like they were necessary enough for me to add them.

onto the review of the pr, the implementation is pretty similar to the plugins mentions though there would need to be some changes before it would be accepted. The first is removal of the new database table. While it's performance is probably better, the purpose of client preferences is exactly for this kind of thing. I'm interested if anyone else has comments on whether it should be a new table or a client preference.

Other than a few minor tweaks the other big concern is creating a native in shavit-core to fire a forward that shavit-zones listens to in order to communicate between the two. I can think of one reason why you would do it this way, but I'm curious on what your actual reason is.

kidfearless:
Also not a big deal for simpler features like this, but in the future ideally PR new features into very_good_yes so that if there is an issue with it that it can be sorted out before becoming official.

PR #997, opened 2020-12-12
fix m_flLaggedMovementValue

Nairdaa:
This edit will fix the speed problems, but can (or will) cause issues with TAS plugins. You gotta handle that in your future TAS plugins yourself.

Thanks to the genius himself, GAMMACASE.

### DIFF
--- a/addons/sourcemod/scripting/shavit-core.sp
+++ b/addons/sourcemod/scripting/shavit-core.sp
@@ -1358,7 +1358,17 @@ void VelocityChanges(int data)
 	{
 		return;
 	}
+	/*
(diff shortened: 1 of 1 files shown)


kidfearless:
This will just make it so that other styles get messed up when they enter the zones

Nairdaa:
> This will just make it so that other styles get messed up when they enter the zones

Other styles that are not a part of the timer and should be handled in the external plugins. Anyway, adding another setting in styles config is also an option, true.

kidfearless:
Slow motion is part of the timer, by removing that code you would break slow motion on runs on those same maps. The pr would just be trading one broken style for another.

PR #1001, opened 2021-01-08
Delete frames from cpcache on entry deletion

GAMMACASE:
Fixes currently big memory leak related to segmented cps.


kidfearless:
Nice find!

PR #1004, opened 2021-01-14
Persist checkpoint data

deadw1nter:
(#793)


Nairdaa:
Pog champ deadwinter!. That's something that's been missing and wanted in my opinion. Anyway, I guess it should have a cvar to enable/disable in the misc cfg

PR #1012, opened 2021-02-20
add Shavit_GetTimesTeleported()

rtldg:
Untested so far but probably fine, right? Intended to be used by another plugin but maybe the teleport count could be added to the db eventually?

### DIFF
--- a/addons/sourcemod/scripting/include/shavit.inc
+++ b/addons/sourcemod/scripting/include/shavit.inc
@@ -1960,6 +1960,15 @@ native int Shavit_GetCurrentCheckpoint(int client);
  */
 native void Shavit_SetCurrentCheckpoint(int client, int index);
 
+/**
+ * Gets how many times the client has teleported to checkpoints.
+ *
+ * @param client					Client index
+ *
+ * @return							The number of times the client has teleported to checkpoints.
+ */
+native int Shavit_GetTimesTeleported(int client);
+
 /*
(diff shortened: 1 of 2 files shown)


Nairdaa (commented):
Most of people while segmenting/TAS teleport back to the same checkpoint several times at once (like 6-7 times before actually continuing). I don't see how storing this by the timer itself should be a thing. Waste of memory imo.

Prove me wrong.

rtldg:
> Most of people while segmenting/TAS teleport back to the same checkpoint several times at once (like 6-7 times before actually continuing). 

intended for kz times

> I don't see how storing this by the timer itself should be a thing. 

easier

> Waste of memory imo.

not a lot. also don't you use python in your timer?

Nairdaa:
> not a lot. also don't you use python in your timer?

Shut up, bitch XD, it was easily 3-4 years ago. There is no python at all on any servers I maintain and "my timer" is not even a thing anymore.

kidfearless:
Could be used in the future for pro / tp kz runs

Nairdaa:
> Could be used in the future for pro / tp kz runs

yea and as for now its just a useless waste of memory

blankbhop:
can also be used for a sort of limiter to spamming cp's so people dont start teleporting to 50 cp's in one second

hermansimensen:
0.0002 more megabytes of memory used... how will my server survive this?

PR #1016, opened 2021-02-24
Add map name check to DeleteReplay

deadw1nter:
This fixes the problem that when you delete a record on MapA while the current map is MapX and the replay of MapA doesn't get deleted


rtldg:
I'm going to merge this so I can more easily include my changes.

Issue #646, opened 2018-06-16
[ Request ] Avg/Max velocity


Nairdaa:
Can you please add avg. and max velocity for the record into stats?

shavitush:
use case?

PR #1017, opened 2021-02-26
Avg and max velocity

rtldg:
For #646. ~~Adds `native void Shavit_GetAvgMaxVelocity(int client, float &avg, float &max)`.~~
Adds `Shavit_GetAvgVelocity()`, `Shavit_GetMaxVelocity()`, `Shavit_SetAvgVelocity()`, `Shavit_SetMaxVelocity()`.
Adds `float avgvel, float maxvel` to `Shavit_OnWorldRecord`, `Shavit_OnFinish_Post`, `Shavit_OnFinish`, `Shavit_OnWorstRecord`.
Adds avg/max velocity to `cp_cache_t` and `timer_snapshot_t`
Prints avg/max velocity to run-finish message.
The avg/max velocity should probably be stored in the db eventually too... but we'll wait for that


GAMMACASE:
Isn't it better to have 2 different natives tho? Like currently I would need to create dummy variable to get avg or max speed only.

kidfearless:
Why should this be added directly to the timer instead, instead of with a third-party plugin?

Nairdaa:
> Why should this be added directly to the timer instead, instead of with a third-party plugin?

Because it should be a part of the timer and not a third-party plugin.

kidfearless:
> 
> 
> > Why should this be added directly to the timer instead, instead of with a third-party plugin?
> 
> Because it should be a part of the timer and not a third-party plugin.

Because it should isn't a very good reason to add something to the timer that serves no purpose to the timer.

Nairdaa:
> > > Why should this be added directly to the timer instead, instead of with a third-party plugin?
> > 
> > 
> > Because it should be a part of the timer and not a third-party plugin.
> 
> Because it should isn't a very good reason to add something to the timer that serves no purpose to the timer.

It's been a thing in timers since I started to bhop back in like 2012-2013. It's just another useless stat to see, but cool to have, just like the amount of jumps, strafes or sync in the hud.

addons/sourcemod/scripting/include/shavit.inc
```diff
@@ -1840,6 +1840,18 @@ native int Shavit_PrintToChat(int client, const char[] format, any ...);
...
+ * Gets the average and max velocity of a player's jumps.
+ * Velocity is calculated as a 2d vector.
+ * Average calculation: avg += (vel - avg) / jumps
+ *
+ * @param client					Client index
+ * @param avg						Variable to recieve the average velocity
+ * @param max						Variable to receive the max velocity
+ * @noreturn
+ */
+native void Shavit_GetAvgMaxVelocity(int client, float &avg, float &max);
```
kidfearless, line 1853: Create separate getters and setters for each variable

addons/sourcemod/scripting/shavit-core.sp
```diff
@@ -1436,6 +1439,15 @@ void VelocityChanges(int data)
...
 	}
+
+	if (GetTimerStatus(client) == view_as<int>(Timer_Running))
+	{
+		float curVel = SquareRoot(Pow(fAbsVelocity[0], 2.0) + Pow(fAbsVelocity[1], 2.0));
+		float maxVel = gA_Timers[client].fMaxVelocity;
+		gA_Timers[client].fMaxVelocity = (curVel > maxVel) ? curVel : maxVel;
+		// STOLEN from Epic/Disrevoid. Thx :)
+		gA_Timers[client].fAvgVelocity += (curVel - gA_Timers[client].fAvgVelocity) / gA_Timers[client].iJumps;
+	}
```
kidfearless, line 1450: Implement the average as a sum / total frames. That way styles that don't jump can still benefit from an average velocity counter.

Nairdaa:
Did you ever split this into two messages for scroll styles?

rtldg:
yeah it's two messages right now but I never dumped the avg/max velo to the db

PR #1020, opened 2021-02-27
add timestamp to finish-forwards. add Shavit_ShouldSaveReplayCopy and Shavit_OnReplaySaved

rtldg:
some niche api that i'll be using in another plugin
will probably merge in a few hours since it works fine rn unless someone else wants some changes to this

### DIFF
--- a/addons/sourcemod/scripting/shavit-replay.sp
+++ b/addons/sourcemod/scripting/shavit-replay.sp
@@ -130,6 +130,8 @@ bool gB_Late = false;
 Handle gH_OnReplayStart = null;
 Handle gH_OnReplayEnd = null;
 Handle gH_OnReplaysLoaded = null;
+Handle gH_ShouldSaveReplayCopy = null;
+Handle gH_OnReplaySaved = null;
 
 // server specific
 float gF_Tickrate = 0.0;
@@ -244,9 +246,11 @@ public void OnPluginStart()
 	LoadTranslations("shavit-replay.phrases");
 
 	// forwards
-	gH_OnReplayStart = CreateGlobalForward("Shavit_OnReplayStart", ET_Event, Param_Cell, Param_Cell, Param_Cell, Param_Cell);
-	gH_OnReplayEnd = CreateGlobalForward("Shavit_OnReplayEnd", ET_Event, Param_Cell, Param_Cell, Param_Cell, Param_Cell);
+	gH_OnReplayStart = CreateGlobalForward("Shavit_OnReplayStart", ET_Event, Param_Cell);
+	gH_OnReplayEnd = CreateGlobalForward("Shavit_OnReplayEnd", ET_Event, Param_Cell);
 	gH_OnReplaysLoaded = CreateGlobalForward("Shavit_OnReplaysLoaded", ET_Event);
+	gH_ShouldSaveReplayCopy = CreateGlobalForward("Shavit_ShouldSaveReplayCopy", ET_Event, Param_Cell, Param_Cell, Param_Cell, Param_Cell, Param_Cell, Param_Cell, Param_Cell, Param_Cell, Param_Cell, Param_Cell, Param_Cell, Param_Cell, Param_Cell, Param_Cell);
+	gH_OnReplaySaved = CreateGlobalForward("Shavit_OnReplaySaved", ET_Ignore, Param_Cell, Param_Cell, Param_Cell, Param_Cell, Param_Cell, Param_Cell, Param_Cell, Param_Cell, Param_Cell, Param_Cell, Param_Cell, Param_Cell, Param_Cell, Param_Cell, Param_Cell, Param_String);
 
 	// game specific
 	gEV_Type = GetEngineVersion();
@@ -967,6 +971,14 @@ public void OnMapStart()
 		CreateDirectory(gS_ReplayFolder, 511);
 	}
 
+	char sPath[PLATFORM_MAX_PATH];
+	FormatEx(sPath, PLATFORM_MAX_PATH, "%s/copy", gS_ReplayFolder);
+
+	if(!DirExists(sPath))
+	{
+		CreateDirectory(sPath, 511);
+	}
+
 	for(int i = 0; i < gI_Styles; i++)
 	{
 		gI_ReplayTick[i] = -1;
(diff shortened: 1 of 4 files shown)


addons/sourcemod/scripting/shavit-replay.sp
```diff
@@ -1276,74 +1287,110 @@ bool LoadReplay(int style, int track, const char[] path)
...
 		}
 
 		if(++iFramesWritten == FRAMES_PER_WRITE || i == iSize - 1)
 		{
-			fFile.Write(aWriteData, CELLS_PER_FRAME * iFramesWritten, 4);
+			if (saveWR)
+			{
+				fWR.Write(aWriteData, CELLS_PER_FRAME * iFramesWritten, 4);
+			}
+			if (saveCopy)
```
kidfearless, line 1372: add new line after closing brace

addons/sourcemod/scripting/shavit-replay.sp
```diff
@@ -1761,52 +1808,90 @@ public void Shavit_OnLeaveZone(int client, int type, int track, int id, int enti
...
+
 	int iSteamID = GetSteamAccountID(client);
 
 	char sName[MAX_NAME_LENGTH];
 	GetClientName(client, sName, MAX_NAME_LENGTH);
 	ReplaceString(sName, MAX_NAME_LENGTH, "#", "?");
 
-	SaveReplay(style, track, time, iSteamID, sName, gI_PlayerPrerunFrames[client], gA_PlayerFrames[client], gI_PlayerTimerStartFrames[client]);
+	char path[PLATFORM_MAX_PATH];
+	SaveReplay(style, track, time, iSteamID, sName, gI_PlayerPrerunFrames[client], gA_PlayerFrames[client], gI_PlayerTimerStartFrames[client], timestamp, makeCopy, makeWR, path, sizeof(path));
```
kidfearless, line 1873: you added a variable isTooLong but never used it to decide if you should save the replay.

addons/sourcemod/scripting/shavit-replay.sp
```diff
@@ -1761,52 +1808,90 @@ public void Shavit_OnLeaveZone(int client, int type, int track, int id, int enti
...
+	Call_PushCell(sync);
+	Call_PushCell(track);
+	Call_PushCell(oldtime);
+	Call_PushCell(perfs);
+	Call_PushCell(avgvel);
+	Call_PushCell(maxvel);
+	Call_PushCell(timestamp);
+	Call_PushCell(isTooLong);
+	Call_PushCell(isWR);
+	Call_Finish(action);
```
kidfearless, line 1856: you grab the return value of the forward but don't use it.

addons/sourcemod/scripting/shavit-replay.sp
Nairdaa: also the string name is `path` and not `sPath` just like the in the function above

rtldg: It's used `bool makeWR = (isWR && !isTooLong);`

kidfearless:
Also, if all you wanted was a copy of their frames when they finish you can already do that with the existing api.

addons/sourcemod/scripting/shavit-replay.sp
rtldg: it's used `bool makeCopy = (action != Plugin_Continue);`

rtldg:
> Also, if all you wanted was a copy of their frames when they finish you can already do that with the existing api.

Not reliably if the Shavit_OnFinish in shavit-replay runs first and calls ClearFrames()

Nairdaa:
there are also things like `int iFramesWritten = 0;` and `int iSize = playerrecording.Length;` and then `bool isTooLong` instead of `bool bOsTooLong`.

Can we please use the same formatting

Nairdaa:
`bool bBypass = (numParams < 2 || view_as<bool>(GetNativeCell(2)));`

Check shavit-core for reference, that's how it is like everywhere in the timer.

addons/sourcemod/scripting/shavit-replay.sp
kidfearless: be sure to follow the general pattern that the rest of the timer uses. While assigning it to a variable is more verbose it's better to be consistent.
```
	Action result = Plugin_Continue;
	Call_StartForward(gH_Forwards_FinishPre);
	Call_PushCell(client);
	Call_PushArrayEx(snapshot, sizeof(timer_snapshot_t), SM_PARAM_COPYBACK);
	Call_Finish(result);

	if(result != Plugin_Continue && result != Plugin_Changed)
	{
		return;
	}
	```

File: addons/sourcemod/scripting/shavit-replay.sp (hunk 3 above)
Comment (kidfearless) [reply addons/sourcemod/scripting/shavit-replay.sp:1856]: be sure to follow the general pattern that the rest of the timer uses. While assigning it to a variable is more verbose it's better to be consistent.
```
	Action result = Plugin_Continue;
	Call_StartForward(gH_Forwards_FinishPre);
	Call_PushCell(client);
	Call_PushArrayEx(snapshot, sizeof(timer_snapshot_t), SM_PARAM_COPYBACK);
	Call_Finish(result);

	if(result != Plugin_Continue && result != Plugin_Changed)
	{
		return;
	}
	```

kidfearless: Also since it's a forward/event a more fitting naming would be Shavit_OnReplayCopy or Shavit_OnBackupReplay

rtldg: > be sure to follow the general pattern that the rest of the timer uses. While assigning it to a variable is more verbose it's better to be consistent.

It doesn't make it verbose, it makes it obvious what is happening.

> Also since it's a forward/event a more fitting naming would be Shavit_OnReplayCopy or Shavit_OnBackupReplay

Which forward name is this about? Shavit_OnReplaySaved runs for whenever a replay is saved, whether it's a copy or not. And I think Shavit_ShouldSaveReplayCopy is a fine name since it's about whether a copy of a replay should be saved. Those two names you offered don't match the forwards as well.

rtldg:
> there are also things like `int iFramesWritten = 0;` and `int iSize = playerrecording.Length;` and then `bool isTooLong = (gCV_TimeLimit.FloatValue > 0.0 && time > gCV_TimeLimit.FloatValue);` instead of `bool bIsTooLong`.
>Can we please use the same formatting?

Using the same formatting would mean isTooLong & isWR should be lowercased since it seems like all parameters in forwards and definitions use lowercase.

> `bool bBypass = (numParams < 2 || view_as<bool>(GetNativeCell(2)));`
> Check shavit-core for reference, that's how it is like everywhere in the timer.

That example already not consistent and it's not like that everywhere in the timer.

PR #1029, opened 2021-03-17
Update core phrases

BoomShotKapow:
Minor change to the core phrases translation file.


rtldg:
didn't even notice this was messed up because i had updated the translation on my server but forgot to copy the translation over / `ln -s` it

BoomShotKapow:
> didn't even notice this was messed up because i had updated the translation on my server but forgot to copy the translation over / `ln -s` it

Yea, it happens. I did something similar in my previous pull request when I left out the SQL query for creating the startpositions table, since it already existed on my database and I completely forgot to test it from scratch.

PR #1030, opened 2021-03-20
close team menu when player join team

bklol:
when player chose their team but teammenu wont close unless player chose same team twice. fix for csgo


rtldg:
The file was saved with CRLF newlines (Windows style) instead of LF.
Fix that and redo your pull request so that only the changes for the menu are in this pull request. Feel free to close / reopen a new pull request if you need

rtldg:
fixed it for you. I haven't tested the changes yet though

PR #1031, opened 2021-03-20
close team menu when player join team

bklol:
when player chose their team but teammenu wont close unless player chose same team twice. fix for csgo


rtldg:
I found that `player_team` has an event bool named `silent`. Using this and enabling event broadcast seems to solve this
https://github.com/shavitush/bhoptimer/commit/6386577ef4149c3676d79f37e9675e15a7c85518

PR #1032, opened 2021-03-21
Change TraceRay masks to MASK_PLAYERSOLID

GAMMACASE:
Some maps have translucent walls and floors and cursor just wont snap on it and instead snaps to whatever is kilometer behind the translucent wall/floor, which makes it just annoying. This changes TraceRay masks to use MASK_PLAYERSOLID that respects texture translucency and fixes described problem.


rtldg:
do you have any example maps I can test on?

GAMMACASE:
bhop_best_map_even_made_233 try to set finish zone there.

PR #1035, opened 2021-03-21
reduce csgo hud size

Nairdaa:
its too big and looks shit on bigger resolutions

### DIFF
--- a/addons/sourcemod/scripting/shavit-hud.sp
+++ b/addons/sourcemod/scripting/shavit-hud.sp
@@ -1274,7 +1274,7 @@ int AddHUDToBuffer_CSGO(int client, huddata_t data, char[] buffer, int maxlen)
 	if((gI_HUDSettings[client] & HUD_ZONEHUD) > 0 && data.iZoneHUD != ZoneHUD_None)
 	{
 		char sZoneHUD[64];
-		FormatEx(sZoneHUD, 64, "<span class='fontSize-xxl' color='#%06X'>", ((gI_Gradient.r << 16) + (gI_Gradient.g << 8) + (gI_Gradient.b)));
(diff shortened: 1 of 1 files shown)


blankbhop:
Why have it hard coded to begin with, could possibly make it a cvar where server owners can scale the hud to whatever they choose.

rtldg:
why have it hardcoded or as a cvar to begin with, could possibly make it a client setting where clients can scale the hud to whatever they choose.

Nairdaa:
> why have it hardcoded or as a cvar to begin with, could possibly make it a client setting where clients can scale the hud to whatever they choose.

Probably the best idea, but idk how well that would work. I also plan on completely reworking the CSGO hud, because the current one looks dogshit.

rtldg:
> > why have it hardcoded or as a cvar to begin with, could possibly make it a client setting where clients can scale the hud to whatever they choose.
> 
> Probably the best idea, but idk how well that would work. I also plan on completely reworking the CSGO hud, because the current one looks dogshit.

Something like adding an `increase hud size` and `decrease hud size` menu options to the !hud menu which are saved in a cookie is probably a good way to do it. I'll probably test this out in a few hours

Nairdaa:
Those few hours sure take you longer than expected :P

rtldg:
> 
> 
> Those few hours sure take you longer than expected :P

any minute now, surely

rtldg:
actually, I was testing but changing the font size does affect the alignment so I was messing with that and potentially not centering stuff and then I got annoyed that the $ symbol was in the hud so I looked into SFUI_ContractKillStart a lot and couldn't find out a way to do something about it

Nairdaa:
Might or not rework the CSGO hud Soon™.

PR #1036, opened 2021-03-21
update menu drawing time

Nairdaa:
There is no point in setting this to 20 seconds or 5 minutes. Just let them stay until the client decides to shut them down.


rtldg:
sure

PR #1037, opened 2021-03-24
Update shavit-advertisements.cfg

Nairdaa:
add more clarification on what's doable and what's not with the advertisements


Nairdaa:
fuck

PR #1049, opened 2021-05-18
added missing prefix

SaengerItsWar:
to prevent fails for people that are using a prefix 
[shavit-zones.smx] Timer (zones GetStageZone) SQL query failed. Reason: Table 'bhoptimer.mapzones' doesn't exist
Query callback. Number of returned results: 399, Maps added to g_aMapList:284, g_aAllMapsList.Length:287, g_mMapList:286


rtldg:
good catch, thank you

PR #1050, opened 2021-05-19
Increase top left WR hud buffer size

GAMMACASE:
The buffer size of 128 is quite small, as in worst case scenario player name length might be exactly 128 bytes, so it'll clearly won't fit 3 of them + some other text, increasing it to something like 512 bytes should be sufficient. Also here's the real example of the problem:
image

### DIFF
--- a/addons/sourcemod/scripting/shavit-hud.sp
+++ b/addons/sourcemod/scripting/shavit-hud.sp
@@ -1786,17 +1786,17 @@ void UpdateTopLeftHUD(int client, bool wait)
 			char sWRName[MAX_NAME_LENGTH];
 			Shavit_GetWRName(style, sWRName, MAX_NAME_LENGTH, track);
 
-			char sTopLeft[128];
-			FormatEx(sTopLeft, 128, "WR: %s (%s)", sWRTime, sWRName);
+			char sTopLeft[512];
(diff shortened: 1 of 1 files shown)


rtldg:
CSGO is weird. I wonder why they ever decided to up the name buffer to 128 (bytes) since steam names can only be 32 bytes + 1 NUL afaik. But also this seems fine so merging

GAMMACASE:
I'm not entirely sure, but that's 32 4-byte symbols (128 total) that is allowed by steam, not 32 bytes. And that's not really a CSGO problem as you can't change the name in game, it uses your steam name, and cuz cyrillic symbols are 2 byte long they take a lot more space then any other english words.

Nairdaa:
stop using cyrillic, start using human letters, problem solved

rtldg:
> I'm not entirely sure, but that's 32 4-byte symbols (128 total) that is allowed by steam, not 32 bytes.

That doesn't appear to be true. If I use some 4-byte emojis then my steam name can only have 8 total emojis.
I used some 2-byte cryillic characters to test also and only was able to use 16 of them.

GAMMACASE:
Well it actually allows you to set up to 32 symbols, but it'll truncate it when you save it for some reason, anyway, the problem was solved.

PR #1051, opened 2021-05-23
Fix player world damage with shavit_misc_godmode 1

deadw1nter:
just an 1-line fix.
preview: https://streamable.com/kilkhh

edited title cuz nairdaaaaaaaaaa said it


Nairdaa:
1 symbol fix*

rtldg:
good catch. i love these bugs that exist for 6 years and nobody notices

Nairdaa:
> good catch. i love these bugs that exist for 6 years and nobody notices

It was a feature

PR #1052, opened 2021-06-02
Fix infinite loop in RemoveFromString()

GAMMACASE:
Currently it is possible to hit an infinite loop if the sm_ccname has non correct length sequence at the end (eg. ``^FFF``, where it would expect ``^FFFFFF``), this pr adds the necessary checks to verify that end of the string was set correctly.

Also ``when u pr make sure to notice "nairda found a new feature in the timer to crash any server with cc access"`` (c) Nairda.


Nairdaa:
I approve this fix!

rtldg:
well, that fixed that, hopefully more problems don't pop up lol

PR #1067, opened 2021-07-21
ConVar to allow movement/noclip while paused

OliviaMourning:
Added on my server a while ago, players enjoy it and I don't see issues with cheating/exploiting. Just a QoL feature that is disabled by default. Uses the shavit_core_pause_movement cvar to enable.

### DIFF
--- a/addons/sourcemod/scripting/include/shavit.inc
+++ b/addons/sourcemod/scripting/include/shavit.inc
@@ -2153,6 +2153,14 @@ native bool Shavit_IsPaused(int client);
 native int Shavit_CanPause(int client);
 
 /**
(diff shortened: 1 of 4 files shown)


rtldg:
I merged this but a bit edited in the very_good_yes branch https://github.com/shavitush/bhoptimer/commit/af8ab48fc9652cbcaf7fa21ed966d27848d7005b
Main thing changed was to save/restore the savestate when pausing/resuming so it should be a bit harder to abuse triggers and such. Also didn't want to add the `Shavit_PauseMovement` native.
Let me know if there's any problems with that but I'll close this in the meantime since it's merged

PR #1074, opened 2021-08-23
Handle multi byte symbols in player names in !recent menu correctly

GAMMACASE:
Currently there's a problem with multi byte characters being cut early in the !recent menu:
image

Here's after this pr:
image

This pr should handle all multi byte symbols correctly, also there's already a TrimPlayerName() implementation in shavit-hud, and I'm unsure if it's better to just move it to shavit.inc as a stock and use from there as it might come handy again somewhere else idk.

### DIFF
--- a/addons/sourcemod/scripting/include/shavit.inc
+++ b/addons/sourcemod/scripting/include/shavit.inc
@@ -610,6 +610,28 @@ stock float GetAngleDiff(float current, float previous)
 	return diff - 360.0 * RoundToFloor((diff + 180.0) / 360.0);
 }
 
+// https://forums.alliedmods.net/showthread.php?t=216841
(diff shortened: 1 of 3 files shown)


rtldg:
>also there's already a TrimPlayerName() implementation in shavit-hud, and I'm unsure if it's better to just move it to shavit.inc as a stock and use from there as it might come handy again somewhere else idk.

Yeah, moving it to shavit.inc would be a good idea. If you'd do that it'd be awesome. Might be better to name it something like TrimCodepoints or something like that but it's not a big deal if you don't. It would be nice for it to be actual grapheme cluster trimming, but that sounds like a pain to do in sourcepawn.

GAMMACASE:
Alright, moved it shavit.inc, and renamed to TrimDisplayString().

PR #1081, opened 2021-09-29
Fixed maps with a >10 min average having the minimum time limit

Sirhephaestus:
Still not ideal since the average is divided by the minimum amount of times rather than the amount of times that are actually used in calculating the average. Also tweaked numbers a bit and cleaned up other things.


Sirhephaestus:
Honestly the values still need tweaking but the other stuff i changed is still good to go. Most maps have a long limit.

PR #1103, opened 2022-01-02
fix syntax to match the rest of the api

Nairdaa:


rtldg:
the entire file doesn't match and that's okay

PR #1106, opened 2022-01-06
[Prebuilt] Fix start zone in wrong place.

akanora:


Nairdaa:
I mean the mapper wanted it that way, no?

akanora:
There is a start zone where the end zone should be. He may be confused or dumb.

Nairdaa:
Or both, just like most mappers

rtldg:
yeah those are shitty zones

PR #1114, opened 2022-01-14
basic autostrafe

NukoOoOoOoO:


addons/sourcemod/scripting/shavit-tas.sp
```diff
@@ -547,6 +547,17 @@ public Action OnPlayerRunCmd(int client, int& buttons, int& impulse, float vel[3
 		}
 		else if (type == AutostrafeType_Basic)
 		{
+			// Maybe use angle to determine the direction? using mousedx is kinda inaccurate
```
kidfearless, line 550: mouse doesn't work for a lot of players, should really be angle change instead.

PR #1116, opened 2022-01-15
cosmetics

Nairdaa:
ranks are always prefixed with "#", even on btimes and fuckithops copypasta timer


rtldg:
https://github.com/shavitush/bhoptimer/commit/062efd4772ab97782dbf77bb56f546ce7f0e4416

PR #1121, opened 2022-02-09
update mapchooser

Nairdaa:
add default "test_" maps to be excluded from being shown in nominations and mapvotes


Nairdaa:
image
le bad

PR #1123, opened 2022-02-15
Change the behaviour of shavit_misc_resettargetname

GAMMACASE:
This change sets ``shavit_misc_resettargetname`` to 0 being default. Also does changes how events are handled when player is teleported from outside the start zone. This implementation accounts for the OnStartTouch events that might be in the start zone triggers and wont clear these events out.

Main logic behind the implementation is that events are applied at the same tick after teleport from both the trigger that was teleported from (OnEndTouch) and to the destination trigger (OnStartTouch), but they are applied too late for the Shavit_OnStartPre() forward to catch them, thus to clearly erase all events from the OnEndTouch DoPhysicsUntouch() called (forcing events to be applied) and events are cleared right after (This is done for 2 ticks after the teleport was done outside the start zone, this value is fine tuned and was fine in my tests, although I'm not sure if client hacks could delay the packets and cause the difference between ticks here for more than 2, I wasn't successful at it with built in net_* commands or trying to lag the client via snd_restart). After the DoPhysicsUntouch() is called and all the events are cleared for the last time (after 2 ticks), the new OnStartTouch would be triggered for the start zone triggers which after they would apply the expected events.

### DIFF
--- a/addons/sourcemod/scripting/shavit-misc.sp
+++ b/addons/sourcemod/scripting/shavit-misc.sp
@@ -2168,6 +2168,32 @@ public Action Shavit_OnStartPre(int client)
 	if((gI_LatestTeleportTick[client] <= curr_tick <= gI_LatestTeleportTick[client] + 1) && 
 		!gB_WasInStartZoneBeforeTeleport[client] && curr_tick != tick_served[client])
 	{
+		if(gCV_ResetTargetname.BoolValue)
+		{
+			char targetname[64];
+			char classname[64];
+			
+			if (Shavit_GetClientTrack(client) == Track_Main)
+			{
+				gCV_ResetTargetnameMain.GetString(targetname, sizeof(targetname));
+				gCV_ResetClassnameMain.GetString(classname, sizeof(classname));
+			}
+			else
+			{
+				gCV_ResetTargetnameBonus.GetString(targetname, sizeof(targetname));
+				gCV_ResetClassnameBonus.GetString(classname, sizeof(classname));
+			}
+			
+			DispatchKeyValue(client, "targetname", targetname);
+			
+			if (!classname[0])
+			{
+				classname = "player";
+			}
+			
+			SetEntPropString(client, Prop_Data, "m_iClassname", classname);
+		}
+		
 		MaybeDoPhysicsUntouch(client);
 		ClearClientEvents(client);
 		
@@ -2188,32 +2214,6 @@ public Action Shavit_OnStart(int client)
 	if (gB_Eventqueuefix)
 	{
 		SetClientEventsPaused(client, false);
-	}
-	
-	if(gCV_ResetTargetname.BoolValue)
-	{
-		char targetname[64];
-		char classname[64];
-
-		if (Shavit_GetClientTrack(client) == Track_Main)
-		{
-			gCV_ResetTargetnameMain.GetString(targetname, sizeof(targetname));
-			gCV_ResetClassnameMain.GetString(classname, sizeof(classname));
-		}
-		else
-		{
-			gCV_ResetTargetnameBonus.GetString(targetname, sizeof(targetname));
-			gCV_ResetClassnameBonus.GetString(classname, sizeof(classname));
-		}
-
-		DispatchKeyValue(client, "targetname", targetname);
-
-		if (!classname[0])
-		{
-			classname = "player";
-		}
-
-		SetEntPropString(client, Prop_Data, "m_iClassname", classname);
 	}
 
 	return Plugin_Continue;
--- a/addons/sourcemod/scripting/shavit-misc.sp
+++ b/addons/sourcemod/scripting/shavit-misc.sp
@@ -43,7 +43,6 @@
 #include <shavit/zones>
 #include <eventqueuefix>
 
-#include <shavit/physicsuntouch>
 #include <shavit/weapon-stocks>
 
 #pragma newdecls required
@@ -72,9 +71,6 @@ Function gH_AfterWarningMenu[MAXPLAYERS+1];
 int gI_LastWeaponTick[MAXPLAYERS+1];
 int gI_LastNoclipTick[MAXPLAYERS+1];
 int gI_LastStopInfo[MAXPLAYERS+1];
-
-int gI_LatestTeleportTick[MAXPLAYERS+1];
-bool gB_WasInStartZoneBeforeTeleport[MAXPLAYERS+1];
 
 // cookies
 Handle gH_HideCookie = null;
@@ -136,7 +132,6 @@ Handle gH_Forwards_OnClanTagChangePost = null;
 DynamicHook gH_GetPlayerMaxSpeed = null;
 DynamicHook gH_IsSpawnPointValid = null;
 DynamicDetour gH_CalcPlayerScore = null;
-DynamicHook gH_TeleportDhook = null;
 
 // modules
 bool gB_Checkpoints = false;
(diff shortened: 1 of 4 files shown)


GAMMACASE:
With the latest commit I've added the targetname/classname reset to being performed before the OnStartTouch events are applied from the triggers in a start zone, thus eliminating the whole need in ``shavit_misc_resettargetname`` and ``shavit_misc_resetclassnam`` and other related cvars, as it's done automatically and does work as expected on maps that do not have the reset triggers in place (bhop_solitude as an example) and also on maps that have their own reset triggers (as the events are applied after the plugin targetname/classname reset)

rtldg:
`bhop_overthinker`'s bonus is back to being abusable. Restarting on `bhop_japan` is back to sucking. Restarting on `bhop_drop` is back to sucking, although that map also sucks with eventqueuefix because OnStartTouch boosters are iffy with eventqueuefix. Restarting on `bhop_crash_egypt` is back to sucking. Restarting on `bhop_shutdown` is back to sucking. Restarting on `bhop_space` is back to sucking. Restarting on `bhop_interloper` is really really awful, although I'm not sure if that needed something in the mapfixes. Restarting on `bhop_apathy` is sucks although mapfixes didn't have anything for that before either.

I think there's some more but I'm kind of done checking for now. Let me know if you resolve those

Nairdaa:
touché

rtldg:
rename `shavit_misc_resettargetname` to something like `shavit_misc_resettargetnamexx` so the new default value propagates. resolve the branch conflicts (if that option shows up for you). i'll test more soon

GAMMACASE:
> rename `shavit_misc_resettargetname` to something like `shavit_misc_resettargetnamexx` so the new default value propagates. resolve the branch conflicts (if that option shows up for you). i'll test more soon

I thought about it quite a while, and even thought on completely removing the cvar ``shavit_misc_resettargetname`` completely, but on the other hand having it on 1 would just make some maps that are still not found (like crash_egypt) to be non playable (which would be noticed right away and I assume reported instantly) and that's all that there would be (as if the map has the reset trigger on spawn, the targetname and classname from it would be still applied after the timer reset is done). Also was thinking on changing the ``shavit_misc_resettargetname_main``, etc to some non default value by default, so it would help with some maps and eliminate their existence in the mapfixes config (bhop_crash_egypt, bhop_apathy, bhop_interlopper, bhop_space, maybe others too), and leave only a couple that actually needs a mapfix (the ones that require some specific targetname or classname to be set, bhop_drop (although that map could be fixed by using !addspawn in the trigger above the start), bhop_shutdown, etc), but that might need a cvar renaming tho.

Nairdaa:
updated both shavit-misc and shavit-mapfixes
shavit_misc_resettargetname is set to 0.

**Tested:**
- bhop_japan works
- bhop_crash_egypt works
- bhop_drop works
- bhop_shutdown works
- bhop_interloper is still awkward because of that shit overlay TinyBanana added. Altho, once that overlay ends, map works as intended. You can go spec and back to game without having that overlay again. Another entity removal is required I guess.
- bhop_space works
- bhop_apathy works, no more dumb spawn
- bhop_overthinker - sorry, I don't know the shortcut you're talking about.

GAMMACASE:
Alright, decided not to remove the cvar and currently ended up just renaming it. Anyway that's a minor stuff and could be changed whenever.

addons/sourcemod/scripting/shavit-misc.sp
```diff
@@ -2096,14 +2132,20 @@ public Action Command_Specs(int client, int args)
...
-	if (client > 0 && gB_Eventqueuefix)
+	if(!IsValidEntity(pThis) || !IsClientInGame(pThis) || IsFakeClient(pThis))
 	{
-		ClearClientEvents(client);
+		return MRES_Ignored;
 	}
+	
+	if(!hParams.IsNull(1))
+	{
+		gB_WasInStartZoneBeforeTeleport[pThis] = Shavit_InsideZone(pThis, Zone_Start, -1);
```
rtldg, line 2144: it's possible to stand inside a bonus start zone and then restart to the main track which will cause the event clearing to not happen (and no targetname reset too on maps if enabled). might not introduce too many problems. not sure what the best way to deal with that would be but `Shavit_OnRestart` can be used I guess if you want to try tracking teleporting through restart.

Also, using `!sp` to have a spawnpoint outside of the startzone would also end up with the events & targetname not being reset/set/cleared/etc.

addons/sourcemod/scripting/shavit-misc.sp
```diff
@@ -2112,7 +2154,58 @@ public Action Shavit_OnStartPre(int client)
...
+			if (!classname[0])
+			{
+				classname = "player";
+			}
+			
+			SetEntPropString(client, Prop_Data, "m_iClassname", classname);
+		}
+		
+		MaybeDoPhysicsUntouch(client);
+		ClearClientEvents(client);
```
rtldg, line 2198: needs to be wrapped in `if (gB_Eventqueuefix) { ClearClientEvents(client); }`

addons/sourcemod/scripting/shavit-misc.sp
GAMMACASE: Not sure how the ``!sp`` outside the start zone could be abused tho? Any examples?

rtldg: Not necessarily sure it could be abused unless the map fails to set targetname at the start. The bonus one too of course might be hard to get an example of it breaking specific maps because it depends on bonus placement & potential boosters that overlap or triggers with delays you could reach the bonus start before it goes off.

GAMMACASE: Nah, I get that problem about bonus and main starts, would play around with it. But didn't get the ``!sp`` problem, isn't it like a cp anyway, and if the map don't have any reset triggers on start simply doing ``!r`` would still cause issues if no reset is done.

rtldg: >isn't it like a cp anyway

it's just `sm_r -> Shavit_RestartTimer(client) (Shavit_StopTimer()) -> shavit-zones::Shavit_OnRestart() -> TeleportEntity() [-> Shavit_StartTimer() when not !sp]`.

>if the map don't have any reset triggers on start simply doing `!r` would still cause issues if no reset is done.

yeah, this probably only ends up being a problem on the like 1 or 2 maps that need the targetname reset which wouldn't run since the shavit_misc_forcetargetnamereset code only happens in the first-tick-since-teleport section

GAMMACASE: But when they would enter the start zone the reset would fire again after that anyway (if cvar is enabled that is), as the code is always run when you just stand on the ground in a start zone.

rtldg: Doesn't the targetname reset only happen within the teleported tick +0 / +1 in this?
```
if((gI_LatestTeleportTick[client] <= curr_tick <= gI_LatestTeleportTick[client] + 1) && 
		!gB_WasInStartZoneBeforeTeleport[client] && curr_tick != tick_served[client])
```

GAMMACASE: Yeah, now I see where this is going, you are right, I'll check this too then.

GAMMACASE: Removed the start zone tracking code to prevent abusing the events from bonuses and vise versa. For the !sp problem, the code that resets targetname was moved back to where it was (and now will be fired on every OnStart() forward call).

addons/sourcemod/scripting/shavit-zones.sp
```diff
@@ -4319,6 +4437,33 @@ public void TouchPost(int entity, int other)
...
+				
+				// GAMMACASE: This prevents further abuses related to external events being ran after you teleport from the trigger, with events setup, outside the start zone into the start zone.
+				// This accounts for the io events that might be set inside the start zone trigger in OnStartTouch and wont reset them!
+				// Logic behind this code is that all events in this chain are not instantly fired, so checking if there were teleport from the outside of a start zone in last couple of ticks
+				// and doing PhysicsRemoveTouchedList() now to trigger all OnEndTouch that should happen at the same tick but later and removing them allows further events from OnStartTouch be separated
+				// and be fired after which is the expected and desired effect.
+				// This also kills all ongoing events that were active on the client prior to the teleportation to start and also resets targetname and classname
+				// before the OnStartTouch from triggers in start zone are run, thus preventing the maps to be abusable if they don't have any reset triggers in place
+				if((gI_LatestTeleportTick[other] <= curr_tick <= gI_LatestTeleportTick[other] + 4) && curr_tick != tick_served[other])
```
rtldg, line 4452: what's the reason for the +4 tick range?

GAMMACASE: Cuz the Touch function might skip few ticks when trigger is untouched, so in reality it needs more ticks to fully retrigger all triggers in start zone, and the 4 here was again fine tuned to be fine on the maps I have tested it. It all depends on the trigger call order and may vary from map to map.

rtldg: I'll try to get this merged in soon. This part is still just something I want to look at more though since +1 seems to work fine

GAMMACASE: It does not on the bhop_overthinker (Only in CSS) and the test map I built, with multiple triggers inside each others. On bhop_overthinker that is what you were getting before I'm assuming, first time teleporting to the bonus there causes the events to not be refired (teleports within the bonus trigger works just fine and events would be applied). Also I don't really see how there additional 3 ticks might cause any issues, so if it adds the additional confidence that events would be refired, I'd leave that in.

rtldg: Ah, yep I see now. Going to do a commit for like two things and then squash and merge this

PR #1128, opened 2022-03-12
Timerban

Nukoooo:
Implemented for nairda

### DIFF
--- a/addons/sourcemod/scripting/include/shavit/core.inc
+++ b/addons/sourcemod/scripting/include/shavit/core.inc
@@ -1390,7 +1390,7 @@ native bool Shavit_ShouldProcessFrame(int client);
 native void Shavit_GotoEnd(int client, int track);
 
 /**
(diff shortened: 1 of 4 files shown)


rtldg:
wondering if we should do anything in this pr for the blacklist style plugin. it might just break that plugin if merged as is.

NukoOoOoOoO:
i blame nairda for requesting this. ask nairda @Nairdaa

Nairdaa:
> wondering if we should do anything in this pr for the blacklist style plugin. it might just break that plugin if merged as is.

who cares about external plugins kek

Nairdaa:
they should edit their plugin to something like "if timerbanned, do timer anyway" and voila

Nairdaa:
Been over a year already 😢

Nukoooo:
not needed and it can be done as a third-party plugin, closing!!!!!!!!

PR #1130, opened 2022-03-17
fix creating replay-directories issue

Ciallo-Ani:
https://discord.com/channels/389675819959844865/934225273866563605/953935648065191976

these implementations should not be done before the style configs loaded

this fix is for those who install bhoptimer at the first time


rtldg:
added some more commits to touch up some other things since the opportunity was here

PR #1134, opened 2022-03-19
Add bhop_nux to mapfixes

BoomShotKapow:


Nairdaa:
boomshot alive ;o

BoomShotKapow:
> boomshot alive ;o

Shh....

GAMMACASE:
What could be abused there via targetnames/classnames?

BoomShotKapow:
> What could be abused there via targetnames/classnames?

To be honest, I just enabled it since the blocks were resetting every time you jumped on them. If there's a proper way to handle that, then I'll close this.

I'm also noticing it on: bhop_culture2

GAMMACASE:
Do you use latest version of the timer? Doesn't seems to happen to me with changes done here https://github.com/shavitush/bhoptimer/pull/1123

rtldg:
this seems to be solved by #1135 and https://github.com/hermansimensen/eventqueue-fix/pull/17 (or at least I don't encounter the reset-on-jump-on-block thing you mentioned with those prs and rngfix)

PR #1137, opened 2022-04-01
merge CP menus & shavit-kz.sp, etc.

sh4hrazad:


Nairdaa:
What about remove that kzcp shit once and for all finally instead?

Nairdaa:
Very nice :)

rtldg:
> note: maybe better to change TAS's default ranking multiplier to "0.0" since segmented style is also "0.0".

true that. i'm going to spoof a commit by you that does that 👀

PR #1139, opened 2022-04-12
Shavit_OnOpenSubMenu

xRz0:


Ciallo-Ani:
whose submenu

rtldg:
WR sub menu. I already talked to them about renaming but I haven't added a commit for that yet. The goal for this was to add extra info to a run's page (like the number of wallboosts from a parkour style for example).
I kind of want to add a JSON column to playertimes that the plugins put random info into and then pass that json to callbacks like this so it can be used so just haven't added commits to this yet

PR #1158, opened 2022-07-24
Use signature for m_surfaceFriction instead of doing netprop±offset

NukoOoOoOoO:
No tf2 because I hate that game.
The method used for csgo can be also used for css/tf2, but considering vtable are exposed so I chose to use [CGameMovement::CategorizePosition](https://github.com/ValveSoftware/source-sdk-2013/blob/0d8dceea4310fde5706b3ce1c70609d72a38efdf/mp/src/game/shared/gamemovement.cpp#L3774-L3780)

### DIFF
--- a/addons/sourcemod/gamedata/shavit.games.txt
+++ b/addons/sourcemod/gamedata/shavit.games.txt
@@ -44,6 +44,14 @@
 
 	"csgo"
 	{
+		"Addresses"
+		{
+			"m_surfaceFriction"
+			{
+				"signature" "CBasePlayer->m_surfaceFriction"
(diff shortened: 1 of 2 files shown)


rtldg:
> No tf2 because I hate that game.

TF2 isn't supported by the tas plugin anyway so no problem.
Note to self: `140.0` (00 00 0c 43) is good value to use when for finding `CategorizePosition` on Windows

rtldg:
I should also mention that I didn't test the sigs on anything other than CSS + Windows, so that's just something that I'll have to do before the next release.

PR #1160, opened 2022-07-24
[TF2] Fix broken NoWeaponDrops (shavit-misc)

### LINKED ISSUE #1159: [TF2] shavit-misc's NoWeaponDrops is broken
You can't purposefully drop items in TF2 like you can in CS:S/GO (no `drop` command). However, when you die (e.g. `kill`), your current weapon is dropped for other team members to use along with an ammo pack. The big difference in TF2 though, is that this dropped weapon becomes a new **tf_dropped_weapon** entity and therefore never satisfies the `FindEntityByClassname(ent, "weapon_*")` check in shavit-misc's Timer_Cron when gCV_NoWeaponDrops.BoolValue is true.
https://github.com/shavitush/bhoptimer/blob/68a57b912341b9f02e8ce3b48b05773b58771d8c/addons/sourcemod/scripting/shavit-misc.sp#L959-L970

SDKHook_WeaponDrop also doesn't behave as you would expect in TF2. When you die, the WeaponDrop callback contains the entity index of the last equipped weapon but NOT the entity index of the newly created **tf_dropped_weapon**. Unless the player changes their loadout, the weapon entity index remains the same before and after respawning, so the only thing  OnWeaponDrop does is kill whatever weapon entity was equipped at time of death.
https://github.com/shavitush/bhoptimer/blob/68a57b912341b9f02e8ce3b48b05773b58771d8c/addons/sourcemod/scripting/shavit-misc.sp#L1474-L1480

So basically neither Timer_Cron or the WeaponDrop callback kill dropped weapons in TF2. Luckily though TF2 has a built-in ConVar for this functionality, ` tf_dropped_weapon_lifetime`, so there’s no need to use a timer or hook WeaponDrop.

jedso:
Fixes #1159 by using the built-in TF2 ConVar ` tf_dropped_weapon_lifetime` and killing all **tf_dropped_weapon** entities when shavit_misc_noweapondrops is changed to 1.

The TF2_KillDroppedWeapons function is necessary because existing dropped weapons continue to live out their lifetime even after `tf_dropped_weapon_lifetime` is changed to 0.

### DIFF
--- a/addons/sourcemod/scripting/shavit-misc.sp
+++ b/addons/sourcemod/scripting/shavit-misc.sp
@@ -295,7 +295,7 @@ public void OnPluginStart()
 
 	mp_humanteam = FindConVar((gEV_Type == Engine_TF2) ? "mp_humans_must_join_team" : "mp_humanteam");
 	sv_disable_radar = FindConVar("sv_disable_radar");
-	if (gEV_Type == Engine_TF2) tf_dropped_weapon_lifetime = FindConVar("tf_dropped_weapon_lifetime");
+	tf_dropped_weapon_lifetime = FindConVar("tf_dropped_weapon_lifetime");
 
 	// crons
 	CreateTimer(10.0, Timer_Cron, 0, TIMER_REPEAT);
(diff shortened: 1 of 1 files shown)


rtldg:
Thank you for the pr :)

PR #1164, opened 2022-07-27
Update README.md

Nairdaa:
just in case


rtldg:
don't open a million single-edit PRs to readme please. just make one PR and edit the files from it and that'd make me super super happy.

This change (`RECOMPILE ALL YOUR PLUGINS THAT USE `#include <shavit>` WITH SM _1.10_ OR STUFF WILL BREAK`) isn't applicable so it's not going to be added. I'm going probably merge one of the other PRs with different wording

PR #1170, opened 2022-09-13
Feat/autorestart

MSWS:
Similar to SurfTimer's auto reset, this MR adds the `sm_autorestart` command to the BHop timer.
Players that have this enabled will automatically be reset if their current run time exceeds their PB.


rtldg:
I made some small changes for style & variable names. Also removed the `ProcessCookie` function and the `SetClientCookie(client, cookie, "0")` since this was a good opportunity to remove that since it's not needed.

PR #1171, opened 2022-09-14
Add replay route preview

MSWS:
Heavily inspired (and copied) from [KZGlobalTimer](https://bitbucket.org/kztimerglobalteam/kztimerglobal/), this adds tiny little dots along the route that the Normal Main track replay bot takes. Very useful for knowing where to go.

- Adds `sm_replayroute` command for players to enable/disable
- Adds `shavit_replay_route` cookie to persiste option (defaults to true)

Requires server owners to update their translation files.

### DIFF
--- a/addons/sourcemod/scripting/shavit-replay-route.sp
+++ b/addons/sourcemod/scripting/shavit-replay-route.sp
@@ -1,0 +1,245 @@
+/*
+ * shavit's Timer - Replay Bot Routing
+ *
+ * This file is part of shavit's Timer (https://github.com/shavitush/bhoptimer)
+ *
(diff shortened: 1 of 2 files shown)


rtldg:
I don't really want to merge this as is. It does work, but it doesn't show a route as prettily as some other replay route/ghost plugins I've seen (arcing lines / squares at jump positions / etc).
I wasn't sure what I wanted to do for this so I slacked on commenting until now.
So this version of the plugin imo should stay as a thirdparty plugin / outside of this repo.
If you want to close the PR or keep it open to work on it, either way is fine of course.

PR #1174, opened 2022-09-24
show timer in hud when started in start zone 

lilac1337:
aaaaaaaaaaaaaaaa


rtldg:
I'm probably going to remove the cvar checking and make it only check if the time < 0.05.
shavit-misc does a bit more than just checking the cvar right now for example
https://github.com/shavitush/bhoptimer/blob/87c2b1e4368b7553ff5495a681dd566c54c57732/addons/sourcemod/scripting/shavit-misc.sp#L1320-L1327

rtldg:
Removed the convar stuff due to reasons in the above comment and also changed the time it compares against from `0.05s` to `0.3s`. Something to do with teleporting/restarting causes it to count up to `0.2s` sometimes before the timer restarts further

PR #1176, opened 2022-09-26
 Add migration to update or add auth FK constraint in playertimes

### LINKED ISSUE #1175: [Bug] playertimes table has FK constraint in v2.5.5a -> v3.0.8, but none >=v3.1.0 (with no migration)
I'm working on a PR for normalizing stage times and in the process enabled foreign key constraints for SQLite with `PRAGMA foreign_keys = ON` (it's off by default). But in doing so it exposed a pretty hilarious bug for SQLite databases:
https://github.com/shavitush/bhoptimer/blob/87c2b1e4368b7553ff5495a681dd566c54c57732/addons/sourcemod/scripting/shavit-core.sp#L2716-L2721
This query which runs OnClientAuthorized REPLACEs the whole user row with an implicit DELETE and INSERT. This is a problem because since v2.5.5a (bb74e14bcfc3285f71757b2fd674cb6754f7f041), auth in the playertimes table has referenced auth in the users table as a FK with ON UPDATE CASCADE ON DELETE CASCADE referential actions.

So basically, with foreign key constraints enabled in SQLite, all playertimes are deleted for the player that joins the server after the user table is updated... not good.

This bug is obviously fixable by just doing another query (not using REPLACE), but this bug doesn't happen on databases created after v3.0.8. For SQLite, it's because the FK constraint was removed from the CREATE TABLE statement (https://github.com/shavitush/bhoptimer/commit/7ca863f04b9319e832c61d000021aa7e7933a356) and for MySQL, because of this commit (60c614df9d95095e1529785a3fc40d97f80d5c13) which added the #if 0 directive around the ADD CONSTRAINT query:
https://github.com/shavitush/bhoptimer/blob/87c2b1e4368b7553ff5495a681dd566c54c57732/addons/sourcemod/scripting/include/shavit/sql-create-tables-and-migrations.sp#L554-L559

So databases created >= v3.1.0 don't have this auth FK in playertimes, but there was no migration for dbs created <= v3.0.8 to remove the existing auth FK creating an arbitrary schema difference depending on which version of the plugin was used to create the db.

**TL;DR:** there should be a migration that either adds the auth FK back to playertimes for dbs created >= v3.1.0 (imo the better option), or removes it if it exists (<= v3.0.8).

jedso:
Fixes #1175. I decided to go with `ON UPDATE RESTRICT ON DELETE RESTRICT` referential actions instead of the CASCADE actions used by the constraint in DBs created <= v3.0.8. RESTRICT would have stopped the REPLACE SQLite query from working/cascading to delete all of the user's playertimes. It does mean that all child playertimes have to be deleted before the user table row can be, but this is already how DeleteRestOfUser works so should be fine.

MySQL supports `ALTER TABLE ... ADD CONSTRAINT ... FOREIGN KEY`. Unfortunately, [SQLite does not](https://stackoverflow.com/a/1884841), which means creating a temporary table to store existing playertimes data, dropping playertimes and re-creating playertimes with the correct FK constraint.

For SQLite I initially tried `PRAGMA foreign_key_list(playertimes)` to test for the existence of FKs on playertimes, but the code was starting to get ugly so I switched to using sqlite_master and checking the SQL used to create the table. This meant using StrContains to check for <= v3.0.8 and >= v3.1.0 DBs which isn't exactly elegant (but it works).

I tested this PR with a MySQL DB created at v3.0.8, v3.3.2 and a fresh DB with latest commits. Same for SQLite (with playertimes having >50k rows). Some more testing would definitely be good but should be working as-is.

Only concern I had was that this migration starts immediately after the first 9000 rows of Migration_DeprecateExactTimeInt, and for SQLite this means dropping playertimes and replacing the table without the exact_time_int column. Should be fine though since the id, exact_time_int combinations are still stored in DBResultSet results. If SQL_Migration_DeprecateExactTimeInt_Query fails for whatever reason though, it could mean loss of exact_time_int and the ability to redo Migration_DeprecateExactTimeInt. Probably should encourage SQLite users to backup before this migration.

### DIFF
--- a/addons/sourcemod/scripting/include/shavit/sql-create-tables-and-migrations.sp
+++ b/addons/sourcemod/scripting/include/shavit/sql-create-tables-and-migrations.sp
@@ -59,6 +59,7 @@ static Database gH_SQL;
 static int gI_Driver;
 static char gS_SQLPrefix[32];
 
+char SQLitePTQuery[1024]; // used in Migration_AddPlayertimesAuthFK if db created <= v3.3.2
 int gI_MigrationsRequired;
 int gI_MigrationsFinished;
 
@@ -168,6 +169,7 @@ public void SQL_CreateTables(Database hSQL, const char[] prefix, int driver)
 		FormatEx(sQuery, sizeof(sQuery),
 			"CREATE TABLE IF NOT EXISTS `%splayertimes` (`id` INTEGER PRIMARY KEY, `style` TINYINT NOT NULL DEFAULT 0, `track` TINYINT NOT NULL DEFAULT 0, `time` FLOAT NOT NULL, `auth` INT NOT NULL, `map` VARCHAR(255) NOT NULL, `points` FLOAT NOT NULL DEFAULT 0, `jumps` INT, `date` INT, `strafes` INT, `sync` FLOAT, `perfs` FLOAT DEFAULT 0, `completions` SMALLINT DEFAULT 1, CONSTRAINT `%spt_auth` FOREIGN KEY (`auth`) REFERENCES `%susers` (`auth`) ON UPDATE RESTRICT ON DELETE RESTRICT);",
 			gS_SQLPrefix, gS_SQLPrefix, gS_SQLPrefix);
+		strcopy(SQLitePTQuery, sizeof(SQLitePTQuery), sQuery);
 	}
 
 	AddQueryLog(trans, sQuery);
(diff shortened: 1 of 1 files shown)


rtldg:
> Only concern I had was that this migration starts immediately after the first 9000 rows of Migration_DeprecateExactTimeInt

I'm seeing what I can do to fix this

rtldg:
Added a commit to put missing user auths into the users table (which my db had which made it fail when adding the foreign key).
Seems to work though so I'll merge this

PR #1177, opened 2022-10-01
Add SQLite migration to fix mapzones having NULL id for all rows

jedso:
[SQLite uses ROWID](https://sqlite.org/lang_createtable.html#rowid) to uniquely identify each row in a table, and a column with type INTEGER PRIMARY KEY becomes an alias for that ROWID value. The CREATE TABLE statement for mapzones since the beginning (1.1b) has been `id INT AUTO_INCREMENT`, but the column _has_ to be of INTEGER type for it to act as a ROWID alias (not INT). Because id is therefore acting as a ordinary integer table column, all id rows in mapzones have been NULL in SQLite DBs since the plugin was released.

This migration fixes that and gives each map zone a unique identifier for SQLite DBs.

### DIFF
--- a/addons/sourcemod/scripting/include/shavit/sql-create-tables-and-migrations.sp
+++ b/addons/sourcemod/scripting/include/shavit/sql-create-tables-and-migrations.sp
@@ -60,6 +60,7 @@ static int gI_Driver;
 static char gS_SQLPrefix[32];
 
 bool gB_MigrationsApplied[255];
+char SQLiteMapzonesQuery[1024]; // used in Migration_FixSQLiteMapzonesROWID if db created <= v3.3.2
 
 public void RunOnDatabaseLoadedForward()
 {
@@ -218,6 +219,7 @@ public void SQL_CreateTables(Database hSQL, const char[] prefix, int driver)
 			"CREATE TABLE IF NOT EXISTS `%smapzones` (`id` INTEGER PRIMARY KEY, `map` VARCHAR(255) NOT NULL, `type` INT, `corner1_x` FLOAT, `corner1_y` FLOAT, `corner1_z` FLOAT, `corner2_x` FLOAT, `corner2_y` FLOAT, `corner2_z` FLOAT, `destination_x` FLOAT NOT NULL DEFAULT 0, `destination_y` FLOAT NOT NULL DEFAULT 0, `destination_z` FLOAT NOT NULL DEFAULT 0, `track` INT NOT NULL DEFAULT 0, `flags` INT NOT NULL DEFAULT 0, `data` INT NOT NULL DEFAULT 0, `form` TINYINT, `target` VARCHAR(63));",
 			gS_SQLPrefix);
 		AddQueryLog(trans, sQuery);
+		strcopy(SQLiteMapzonesQuery, sizeof(SQLiteMapzonesQuery), sQuery);
 	}
 
 	FormatEx(sQuery, sizeof(sQuery),
--- a/addons/sourcemod/scripting/include/shavit/sql-create-tables-and-migrations.sp
+++ b/addons/sourcemod/scripting/include/shavit/sql-create-tables-and-migrations.sp
@@ -60,8 +60,8 @@ static Database gH_SQL;
 static int gI_Driver;
 static char gS_SQLPrefix[32];
 
-char SQLitePTQuery[1024]; // used in Migration_AddPlayertimesAuthFK if db created <= v3.3.2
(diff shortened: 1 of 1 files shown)


jedso:
Just had a thought that SQLite DBs created prior to this PR with an _empty_ mapzones table will always return 0 for `SELECT EXISTS(SELECT 1 FROM %smapzones WHERE id IS NULL);` regardless of whether the table is using id INT AUTO_INCREMENT, so it's probably better to use the `SELECT sql FROM sqlite_master` + `StrContains` approach from #1176.

addons/sourcemod/scripting/include/shavit/sql-create-tables-and-migrations.sp
```diff
@@ -540,6 +554,65 @@ public void Trans_DeprecateExactTimeIntFailed(Database db, ArrayStack stack, int
...
+void ApplyMigration_FixSQLiteMapzonesROWID()
+{
+	if (gI_Driver != Driver_sqlite)
+	{
+		InsertMigration(Migration_FixSQLiteMapzonesROWID);
+		return;
+	}
+
+	char sQuery[256];
+	FormatEx(sQuery, sizeof(sQuery), "SELECT EXISTS(SELECT 1 FROM `%smapzones` WHERE id IS NULL);", gS_SQLPrefix);
```
rtldg, line 566: Do you think it would be fine to replace this with `SELECT 1 FROM mapzones WHERE id IS NULL LIMIT 1;`?

addons/sourcemod/scripting/include/shavit/sql-create-tables-and-migrations.sp
```diff
@@ -540,6 +554,65 @@ public void Trans_DeprecateExactTimeIntFailed(Database db, ArrayStack stack, int
...
+}
+
+public void SQL_FixSQLiteMapzonesROWID_Callback(Database db, DBResultSet results, const char[] error, any data)
+{
+	if (results == null)
+	{
+		LogError("Timer error! SQLiteMapzonesROWID migration failed. Reason: %s", error);
+		return;
+	}
+	else if (!results.FetchInt(0)) // No NULL ids
```
rtldg, line 578: This needs `results.FetchRow()` to be called first.
In fact, since `.FetchRow()` returns a bool if a row was fetched, it should be fine to swap it to this:
```suggestion
	else if (!results.FetchRow()) // No NULL ids
```

rtldg:
I'm not a fan of doing the `char SQLiteMapzonesQuery[1024];` global variable thing but that's probably the easiest way that won't require changing much else like wrapping things in functions or whatever.
Also not that it really matters, but I think it's funny to realize that the `bool gB_MigrationsApplied[255];` uses the same amount of memory since bools are cell sized while chars are byte-sized but packed into cell-sized divisible memory.

addons/sourcemod/scripting/include/shavit/sql-create-tables-and-migrations.sp
jedso: Like the original query, this doesn't handle the scenario I mentioned above (empty mapzones table with the old table creation statement). So ideally you'd need to directly test id's type to definitively know if the migration is required or not.

You could theoretically replace the query using a [PRAGMA function](https://www.sqlite.org/pragma.html), but it is listed as experimental so might break in the future:
```suggestion
	FormatEx(sQuery, sizeof(sQuery), "SELECT 1 FROM pragma_table_info('%smapzones') WHERE name = 'id' AND type = 'INT AUTO_INCREMENT' LIMIT 1;", gS_SQLPrefix);
```
This is the only way of accessing PRAGMA results with a SELECT statement (i.e. `SELECT * FROM PRAGMA table_info(mapzones) WHERE ...;` doesn't work). Then again, the docs do say that "specific pragma statements may be removed and others added in future releases of SQLite. There is no guarantee of backwards compatibility" so probably avoiding PRAGMA statements and functions is the best move long term.

Which means, as far as I can tell, the best way of testing id's type is to use `SELECT sql FROM sqlite_master` + `StrContains` as I did in #1176. Let me know if you agree with that approach and I'll commit it.

rtldg:
I resolved the conflicts on the branch since I just merged the other fk pr. So note to self: this migration number is 29.

addons/sourcemod/scripting/include/shavit/sql-create-tables-and-migrations.sp
jedso: OK, done. Should be good now.

PR #1179, opened 2022-11-06
[TF2] Add correct bounding box values for shavit_zones_box_offset

jedso:
TF2's bounding box is `48x48x82` compared to CS:S `32x32x62` and CS:GO `32x32x72`. As such, the default value of gCV_BoxOffset (bbox width / 2) is slightly off the player's center for TF2.

In SetZoneMinsMaxs, the `height` variable is half the bbox height but does not use the correct value for TF2 either (uses CS:S 62.0 instead of 82.0). This PR adds the correct bbox values for TF2.

As an aside, I did wonder if having `height` be constantly set to half the bbox height is really the desired behavior. E.g. if `shavit_zones_box_offset = 0`, why should distance_z be offset by half the bbox height yet have no offset for distance_x/y?

https://github.com/shavitush/bhoptimer/blob/a6da4358de06937b095dcafd79050e4ea937844a/addons/sourcemod/scripting/shavit-zones.sp#L5104-L5114

### DIFF
--- a/addons/sourcemod/scripting/shavit-zones.sp
+++ b/addons/sourcemod/scripting/shavit-zones.sp
@@ -333,9 +333,14 @@ public void OnPluginStart()
 	gCV_Height = new Convar("shavit_zones_height", "128.0", "Height to use for the start zone.", 0, true, 0.0, false);
 	gCV_Offset = new Convar("shavit_zones_offset", "1.0", "When calculating a zone's *VISUAL* box, by how many units, should we scale it to the center?\n0.0 - no downscaling. Values above 0 will scale it inward and negative numbers will scale it outwards.\nAdjust this value if the zones clip into walls.");
 	gCV_EnforceTracks = new Convar("shavit_zones_enforcetracks", "1", "Enforce zone tracks upon entry?\n0 - allow every zone except for start/end to affect users on every zone.\n1 - require the user's track to match the zone's track.", 0, true, 0.0, true, 1.0);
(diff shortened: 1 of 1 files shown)


rtldg:
> As an aside, I did wonder if having `height` be constantly set to half the bbox height is really the desired behavior. E.g. if `shavit_zones_box_offset = 0`, why should distance_z be offset by half the bbox height yet have no offset for distance_x/y?

Yeah, I think you're right about it not being desired/expected/optimal with `shavit_zones_box_offset = 0`.
I'll add a commit for it to this PR and then merge this in.

rtldg:
> I'll add a commit for it to this PR and then merge this in.

Or should I?
Also I wonder if it'd be better to just make the convar a boolean and hardcode the offset per game...

jedso:
> Also I wonder if it'd be better to just make the convar a boolean and hardcode the offset per game...

Not a bad idea. Guess it depends how much control you want to give people. I guess 3 main solutions for this:

1. Boolean convar to toggle between 0 offset and `bbox / 2` like you say (for most people this should be fine)
2. Height offset scales relative to the width offset (something like `(shavit_zones_box_offset / bbox width) * bbox height`). Allows people to set offsets other than 0 or `bbox / 2`
    *  Or alternatively store `shavit_zones_box_offset` as a decimal fraction float where the default value = 0.5, then use that with bbox_width/height to get offsets
3. Separate convars for width/height offset (not sure anybody would want/need that though)

rtldg:
Ended up adding a commit to make it act like a bool & to do the height too.
I'm not going to squash merge it right now so I can stare at it for a bit longer

PR #1182, opened 2022-12-18
Wip sqlite rankings support & GetWeightedPoints deprecation

rtldg:
The sqlite rankings support depends on https://github.com/alliedmodders/sourcemod/pull/1886
For Mysql >= 8.0 & Mariadb >= 10.2, GetWeightedPoints won't be needed anymore.
The queries should be double-checked


rtldg:
The Sourcemod PR got merged, although I didn't pester them further to get it merged into 1.11. Replacing the `dbi.sqlite.ext.dll`/`dbi.sqlite.ext.so` file with the one from the 1.12 releases would probably work though.
I need to add a query to check if `POW` works to this PR & then double-check everything.

rtldg:
I think it works

PR #1187, opened 2023-02-24
late loading improvement and hp reset on restart

xhaooy:


rtldg:
I edited the late-loading cvar flag one to handle the games where the cvars don't exist (and also removed `sv_autobunnyhopping` since the flags on it aren't modified).
I moved the reset hp cvar to shavit-misc too and renamed it to `shavit_misc_restart_with_full_hp`

xhaooy:
oops, my bad with sv_autobunnyhopping cvar...

thanks for accepting :)

PR #1188, opened 2023-04-25
Godmode fix for DMG_CRUSH and point_hurt

azalty:
Fixes DMG_CRUSH (by moving/falling objects for example (with cs_enable_player_physics_box 1))

Handles correctly damage dealt by point_hurt entities (see https://developer.valvesoftware.com/wiki/Point_hurt)

~~**IMPORTANT:** `shavit_misc_godmode 1` will now block damage dealt by ALL entities that aren't players, instead of only 0 (world).
If you don't want to make it like that, please tell me.~~
-- Added `shavit_misc_godmode 4` for that

### DIFF
--- a/addons/sourcemod/scripting/shavit-misc.sp
+++ b/addons/sourcemod/scripting/shavit-misc.sp
@@ -261,7 +261,7 @@ public void OnPluginStart()
 	RegConsoleCmd("sm_adverts", Command_PrintAdverts, "Prints all the adverts to your chat");
 
 	// cvars and stuff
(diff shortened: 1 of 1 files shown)


rtldg:
Can you split these into two separate cvar values (4 & 5)?

azalty:
For the world/fall damage also preventing damage from entities I understand, but what would be the other option?

rtldg:
Oh yeah I guess the thing for case 2 doesn't really need to be split out.

azalty:
This should do, I hope I didn't miss anything and didn't create any error.
Didn't try to compile the plugin.

azalty:
Didn't know `default` had to be in last position. It's apparently not the case in other languages so I thought it would be the same here

rtldg:
yeah it's a sourcepawn quirk that I forgot about too that thankfully the workflow threw an error for

PR #1197, opened 2023-11-11
use correct gravity for LG

lilac1337:
it's historically always been 400


rtldg:
this won't "correct" it for existing timer users but sure

PR #1207, opened 2024-04-24
Add teleport to stage start command

Awesomerly:
On KSF, the defacto CSS surf server, there is a command called /teleport. What it does is take you back to the previous zone you were in without stopping the timer. On bhop the penalty for failing a stage isn't very bad because the ground isn't far away from the player and stages aren't as important to that gamemode, but in surf it can alleviate players from having to fall for 3 seconds just to get back to the stage they were on.

I still need feedback on this.


Awesomerly:
One thing I realized is that there's already a tele command in bhoptimer that teleports to a checkpoint. On KSF /teleport also takes you to the last checkpoint in their similar-but-less-fleshed-out version of the Segmented style

ddorab:
`bind x "sm_tele; sm_teleport"`, unless there's a `tele` command in surf as well that's all they need to do
also I thought it was meant to be an alias for tele but your thing has the same functionality as ksf where it teleports you to the last zone you were in?

Awesomerly:
> I thought it was meant to be an alias for tele

There's only sm_tele, no sm_teleport defined in shavit-checkpoints;

https://github.com/shavitush/bhoptimer/blob/1f1a8979274f64254bfd25cde676ffdd1065749e/addons/sourcemod/scripting/shavit-checkpoints.sp#L188

ddorab:
well sm_tele teleports you to the last checkpoint, so I thought you would do `RegConsoleCmd("sm_teleport", Command_Tele, "Teleports to a checkpoint. Usage: sm_tele [number]"); ` so that players wouldn't have to change their binds but stage zones are not common at all in bhop and players don't really teleport to the last stage anyway like you said. It's up to 🥑to decide whether teleporting to last stage is needed but I just got confused by what you wanted

Awesomerly:
What you said made me think about something; I could make it so that the teleport to last stage command doesn't work in segmented/tas modes, then tell surfers to alias sm_teleport to be sm_tele and sm_stagerestart or w/e

PR #1216, opened 2024-07-30
Add no-jump and autobhop zones.

Awesomerly:
This PR aims to add two types of zones that are utilized in surf maps.
- No-Jump zones kill the player's momentum when they attempt to jump inside the zone. This can be used to patch out cheeses on maps.
- Autobhop zones are self explanatory. When the player is in an autobhop zone, they can auto bhop.

I think this implementation is pretty okay so far but I have a few things to ask about before making it official:
- Is there a better way to implement nojump than what I've done? Two issues I can see with it are that jumping in a nojump zone still adds to the jump tally, and the momentum gets killed in a way to where it looks like the player gets sucked into the ground.
- A lot of maps on KSF use their own naming scheme for hookzones. Would it be worth it to add some kind of backwards compatibility so that already existing maps can have their zones automatically added? Maybe not but it's still worth an ask.
- Is there anything else you would change? Is it possible to reorder the zoning menu so similar items are closer together or is it just in the order that they're specified?


rtldg:
> Is there a better way to implement nojump than what I've done? Two issues I can see with it are that jumping in a nojump zone still adds to the jump tally, and the momentum gets killed in a way to where it looks like the player gets sucked into the ground.

I used the remove-jump-input code from elsewhere instead of setting the absvelocity so I think it's better now.

> A lot of maps on KSF use their own naming scheme for hookzones. Would it be worth it to add some kind of backwards compatibility so that already existing maps can have their zones automatically added? Maybe not but it's still worth an ask.

Automatic zone hooks might not be worth pursuing after all. I've felt that way after realizing how troublesome it is with a zone database like the unfinished zones-cstrike / zones-json stuff. The "orphan-crushing machine" will continue until those revive 😇

> Is there anything else you would change?

I changed a few things mainly for like better if-statement ordering and readability, and a few things for diff related things (trailing commas for late array item).

> Is it possible to reorder the zoning menu so similar items are closer together or is it just in the order that they're specified?

I think it's easiest/best for now to keep it as is / enum-order.


Will merge soon:tm:

PR #1217, opened 2024-08-17
Update replay-stocks.sp

Nairdaa:
fix logic behind testing if it works or not and when to actually throw an error

### DIFF
--- a/addons/sourcemod/scripting/include/shavit/replay-stocks.sp
+++ b/addons/sourcemod/scripting/include/shavit/replay-stocks.sp
@@ -99,3 +99,4 @@ stock void Shavit_Replay_CreateDirectories(const char[] sReplayFolder, int style
 	{
 		CloseHandle(fTest);
 	}
+}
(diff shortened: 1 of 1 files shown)


rtldg:
If `OpenFile` failed then `fTest` would be `null` and `CloseHandle(null)` is okay to do so there's not actually anything broken here although the logic isn't that clear

Nairdaa:
Closing a handle should only be done if it was validly opened. Calling CloseHandle on null may not cause a crash, but it leads to unnecessary function calls. 

The "updated code" fixes the "issue" by ensuring that it checks if the file was successfully opened before attempting to close it.

Nairdaa:
If you spam enough it leads to this: 
image
Can't do anything about ANY of the running replays now, untill this run is finished. Ignore the "unknown" in the name, that's just because it's an old replay and it's missing the player's name in it's data.

kidfearless (approved):
Looks good, merge it

PR #1219, opened 2024-10-14
Startzone speed limit flexibility

Awesomerly:
This PR changes a few things:
- The startzone code in shavit-misc duplicates some functionality for certain zone types, and one of those duplicates doesn't call StyleMaxPrestrafe even though the rest of the code is identical.
- There is now a style configuration flag that overrides fLimit, so modes that use it can enforce a different speedcap than the ~290 that it normally does
- This style configuration can be overridden by the contents of the data field of the startzone. If the startzone has 400 and the config file has 350, it will use 400.
- There is now a new prespeed type (6) that behaves like type 5 but also blocks prehopping. One thing I noticed is that the code to block bhops in mode 2, 3, and 6 only triggers on the first tick that the player hit the ground. You can still scroll and get to a higher speed, which is less of an issue on other styles that start the timer when you go above a certain speed.

Being able to change the zone speed limit like this allows for zones to behave more like surf. In bhop the average prestrafe requires you to jump at the exact last moment while still being on the ground. Surf prestrafes involve jumping up into the air and doing a strafe while falling down past the edge of the startzone. People have argued about whether this technique should be encouraged, but this type of zone is what I and many others are used to. ckSurf has a default of 350 xy speed.

Pitfalls:
- Coding style might not be up to snuff
- This doesn't do anything for stage zones. One thing that makes me feel better about this is that the way that stages are implemented in bhoptimer are really only meant to record splits. ALSO, I checked vibez and it also doesn't do any speed capping on stages so I'm not the only one who decided to not care.

### DIFF
--- a/addons/sourcemod/configs/shavit-styles.cfg
+++ b/addons/sourcemod/configs/shavit-styles.cfg
@@ -35,6 +35,7 @@
 		// Physics
 		"airaccelerate"		"1000.0" // sv_airaccelerate value for the style.
 		"runspeed"			"260.00" // Running speed. Requires DHooks, shavit-misc and shavit_misc_staticprestrafe set to 1.
+		"maxprestrafe"      "0.0" // IDK what description to put for this....
 		"gravity"			"1.0" // Gravity setting, 1.0 for default. Standard for low gravity styles is 0.6.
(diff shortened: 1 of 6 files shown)


Awesomerly:
Just added the new prespeed type.

PR #1222, opened 2025-01-10
Fix database verison parsing

Awesomerly:
In the "DoWeHaveWindowFunctions" function, the timer checks the version numbers by converting it to a float. For some reason, this didn't work for me because sVersion was `10.11.6-MariaDB-0+deb12u1`, then StringToFloat converted that into 10.109999, which isn't greater than the required version of 10.2!

I changed this function to parse the major and minor version and do integer comparisons instead. I haven't tested it thoroughly but it seems to work.


rtldg:
Good catch. It looks like it'll work so a yolo-merge for this too

PR #1231, opened 2025-04-10
Add cvar UnNoclipOnRestart

mourningsickness:
Add functionality to turn off noclip on timer restart, disabled by default u.u

I used to use a separate plugin I made for this, but figured it should be in the main plugin itself as a cvar instead. Since adding as a cvar, may as well see if it is wanted upstream.


mourningsickness:
whitespace may have been messed up, let me fix

PR #1234, opened 2025-04-14
Add FirstLogin column to database and profile

mourningsickness:
# This PR will add a "First Login" stat, complimenting the "Last Login" stat.

# How it is calculated:

## New User
New user joins -> Current time (same as how "Last Login" is set on first connect)

## Migration for existing user:
Has times on server -> Oldest time's timestamp

Does not have times on server -> Last login

Does not have times on server and last login == 0 -> Current time (time of migration)

In my opinion, this is the best way to handle these cases, but feel free to suggest otherwise. Unfortunately we don't have any other metrics to populate existing users' data, but hopefully these are pretty close :^(

The best time to plant a tree is 10 years ago, the second best time is right now.


mourningsickness:
Would sqlite need a separate query using `unixepoch('now')`? I've not used sqlite but am looking and it seems it doesn't have `UNIX_TIMESTAMP()`

mourningsickness:
Added sqlite handling

mourningsickness:
image

Working great!

mourningsickness:
image

New users confirmed working

addons/sourcemod/scripting/include/shavit/sql-create-tables-and-migrations.sp
```diff
@@ -686,6 +689,20 @@ public void Trans_FixSQLiteMapzonesROWID_Error(Database db, any data, int numQue
 	LogError("Timer error! SQLiteMapzonesROWID migration transaction failed. Reason: %s", error);
 }
 
+void ApplyMigration_AddUsersFirstLogin()
+{
+	char sQuery[256];
+	FormatEx(sQuery, sizeof(sQuery), "ALTER TABLE %susers ADD `firstlogin` int(11) DEFAULT -1 NOT NULL %s;", gS_SQLPrefix, (gI_Driver == Driver_mysql) ? "AFTER `lastlogin`" : "");
```
rtldg, line 695: You should do `INT NOT NULL DEFAULT -1` here instead

addons/sourcemod/scripting/include/shavit/sql-create-tables-and-migrations.sp
```diff
@@ -686,6 +689,20 @@ public void Trans_FixSQLiteMapzonesROWID_Error(Database db, any data, int numQue
...
+	QueryLog(gH_SQL, ApplyMigration_AddUsersFirstLogin2222222_Callback, sQuery, Migration_AddUsersFirstLogin, DBPrio_High);
+}
+
+public void ApplyMigration_AddUsersFirstLogin2222222_Callback(Database db, DBResultSet results, const char[] error, any data)
+{
+	char sQuery[256];
+	FormatEx(sQuery, sizeof(sQuery), "UPDATE %susers SET firstlogin = IF((SELECT COUNT(*) FROM %splayertimes WHERE %splayertimes.auth = %susers.auth) > 0, (SELECT MIN(`date`) FROM %splayertimes WHERE auth = %susers.auth), IF((%susers.lastlogin != 0), %susers.lastlogin, %s));", gS_SQLPrefix, gS_SQLPrefix, gS_SQLPrefix, gS_SQLPrefix, gS_SQLPrefix, gS_SQLPrefix, gS_SQLPrefix, gS_SQLPrefix, (gI_Driver == Driver_mysql) ? "UNIX_TIMESTAMP()" : "unixepoch('now')");
```
rtldg, line 702: This isn't compatible with sqlite

rtldg:
merged by https://github.com/shavitush/bhoptimer/commit/3d0cd963195e55e836e71028b28c469d5e4fff0a

PR #1236, opened 2025-04-25
fix: prevent unnecessary jump when using !r or switching to TAS

akanora:


rtldg:
This looks like it breaks the edgejump so I did a different fix https://github.com/shavitush/bhoptimer/commit/7e1c3f0dc31b79a093da1995d40888a890da9cef

PR #1241, opened 2025-05-09
shavit-chat - fix chat messages near (or at) max length overflowing into a broken message

mourningsickness:
image


rtldg:
you should put the `MAXLENGTH_NAME` back to 128 and then set `MAXLENGTH_TEXT` to like 250

mourningsickness:
> you should put the `MAXLENGTH_NAME` back to 128 and then set `MAXLENGTH_TEXT` to like 250

:3

PR #1248, opened 2025-07-05
core.inc - sPrefix[64] -> sPrefix[128]

mourningsickness:
after all, why not allow pretty gradients for server prefixes?

Screenshot from 2025-07-05 02-29-33
Screenshot from 2025-07-05 02-29-08

### DIFF
--- a/addons/sourcemod/scripting/include/shavit/core.inc
+++ b/addons/sourcemod/scripting/include/shavit/core.inc
@@ -98,7 +98,7 @@ enum struct stylestrings_t
 
 enum struct chatstrings_t
 {
-	char sPrefix[64];
+	char sPrefix[128];
 	char sText[16];
 	char sWarning[16];
 	char sVariable[16];
--- a/addons/sourcemod/scripting/shavit-core.sp
+++ b/addons/sourcemod/scripting/shavit-core.sp
@@ -2195,12 +2195,27 @@ public int SemiNative_PrintToChat(int client, int formatParam)
 		//TODO: handle if there is a color code applied at iCut that != sText and apply it to the start of sBuffer2
 		char[] sBuffer1 = new char[maxLen+1];
 		char[] sBuffer2 = new char[maxLen+1];
+		char sLastColor[11];
+
 		strcopy(sBuffer1, maxLen+1, sBuffer);
 
 		int iCut = FindCharInString(sBuffer1, ' ', true);
+		int iColor = FindCharInString(sBuffer1, '\\', true);
+		if(iColor != -1 && StrEqual(sBuffer1[iColor+1], "x") && StrEqual(sBuffer1[iColor+2], "0") && StrEqual(sBuffer1[iColor+3], "7"))
+		{
+			if(iCut - iColor <= 10)
+			{
+				iCut = iColor-1;
+			}
+			else if(!StrContains(sBuffer1[iColor], gS_ChatStrings.sText))
+			{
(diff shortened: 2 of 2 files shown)


mourningsickness:
> I don't think anyone should actually be using more than 64 bytes for a prefix but I'd accept this PR anyway if you want to reopen it. Cutehops does have the same change in their timer fork I think even.

is it possible to have something like this inside the SemiNative_PrintToChat function? not too sure how it would work with the bfmsg kinda thing..

```if(sizeof(message) > maxLen)
    int message1End = message - maxLen;
    int message2Start = message1End+1;
    print(message[:-message1end];
    print(message[message2];```

to allow (theoretically) unlimited Shavit_PrintToChat message lengths, with each line > 1 being printed without the prefix.. so they are pseudo wrapped lines, but really separate messages to get around the limit all together? it would have to check if the split spot is the middle of a color code, etc.. but i can work on this more if it's possible with the bfmsg thing

rtldg:
> to allow (theoretically) unlimited Shavit_PrintToChat message lengths, with each line > 1 being printed without the prefix.. so they are pseudo wrapped lines, but really separate messages to get around the limit all together? it would have to check if the split spot is the middle of a color code, etc.. but i can work on this more if it's possible with the bfmsg thing

Yeah, it'd mean creating a SayText2 message for each chunk/line we send. It'd be good to split between UTF-8 characters too (rip extended grapheme clusters lol) when doing colors splits but luckily those aren't too common.

mourningsickness:
> > to allow (theoretically) unlimited Shavit_PrintToChat message lengths, with each line > 1 being printed without the prefix.. so they are pseudo wrapped lines, but really separate messages to get around the limit all together? it would have to check if the split spot is the middle of a color code, etc.. but i can work on this more if it's possible with the bfmsg thing
> 
> Yeah, it'd mean creating a SayText2 message for each chunk/line we send. It'd be good to split between UTF-8 characters too (rip extended grapheme clusters lol) when doing colors splits but luckily those aren't too common.

gotcha, tbh it didn't need much more even for my use case.. so maybe a _slight_ increase would be good, since it would allow for a bit longer of prefixes, but not cut off messages either. i'll play around with the value and edit. perhaps somewhere around 80

mourningsickness:
i'll work on this some more today. i should be able to finish it now, using `FindCharInString`. i had all of the code ready, except for the reverse parsing so that'll help a ton :pray:

mourningsickness:
btw, i limited the printing to two messages instead of using an array of message strings or something to allow "theoretically" any length message because, realistically, why would it need more than two? and also, if it were 3+, it would just clutter the already small chat box with "one" message.. and tbh i don't think extremely long messages like that should be supported. a 500 character limit is very reasonable i think, given that it is nearly double the previous limit. (not actually double because sBuffer[512] would be > sBuffer1 + sBuffer2 in css.

mourningsickness:
yes, needs more tweaking

image

mourningsickness:
^granted, that is with someone else's custom font/hud, so the line wrapped differently that it would with the vanilla one, but the issue is pictured

mourningsickness:
it would appear to be fixed, but i will continue to test. if this doesn't have any more issues, it should be good because i added previous color handling, so it will use the proper colors on line 2 if it got cut while a color other than sText was in use.

image
image

mourningsickness:
tbh not the prettiest if it does cut, BUT it is a very edge case i'd guess, with code that won't be reached without a looong sPrefix or even longer message. even with kawaii's long gradient, it only cuts very few messages, usually with long usernames

PR #1249, opened 2025-07-07
shavit-misc.sp - don't stop timer on tpto if paused with pause movement enabled

mourningsickness:


rtldg:
> tbh it should have the "Are you sure you want to stop your timer?" confirmation.. but i didn't see where that was called in the plugins. I'd imagine OnStopPre but not sure. That should def be prompted though with the tpto command

Ooo yeah that'd be a good idea.
You should be able to copy this
https://github.com/shavitush/bhoptimer/blob/4f43b628ae1f1f923ceb518528de074b2c765966/addons/sourcemod/scripting/shavit-misc.sp#L2456-L2466
and then do a callback similar to `DoStyleChange` that just calls `Teleport(client, gI_LastStopInfo[client])`

mourningsickness:
perhaps ShouldDisplayStopWarning should be a stock in the includes instead of just in shavit-misc. i'll just copy the logic for now though

mourningsickness:
added, should be good now.

rtldg:
I haven't tested but I'm sure it's fine 😇

PR #34, opened 2016-06-01
Teleport to end

OfficialExtan:

PR #1255, opened 2025-08-01
Add Style Ranks

mourningsickness:
## This PR (finally) adds per-style rankings!

The new !top and !rank menus allow you to pick from an "overall" rank (how bhoptimer ranks work now) or each configured (and not disabled) style.

Functionality is working, with no found errors or bugs. It should be optimized further, but I wanted to open the PR as soon as it was stable to get some insight into how to best do that.

Next on the todo is to incorporate it into shavit-chat with something like {stylerank} variables to be used in chat tags.

Screenshot from 2025-07-31 19-45-25

Screenshot from 2025-07-31 19-45-34

Screenshot from 2025-07-31 19-45-47

Screenshot from 2025-07-31 19-46-03


mourningsickness:
i added this to the bhop server (577k playertimes, 130k users) and performance is actually not bad. i still want to optimize further, but i was expecting it to take a noticeable amount of time for the migration and on each map change.

mourningsickness:
honestly, it doesn't even appear to have performance issues whatsoever.. so it could ship as-is. i'd prefer to get input first though, to make sure there isn't a good way to make it more efficient, say, if there are 100+ styles. My servers only have slightly more than 20 at the moment.

mourningsickness:
Style rank based chatrank support has been added, using {styleXrank} where X is the style ID in shavit-chat.cfg. It works the same as {rank}, being replaced with an int, allowing things like [#{style0rank} Normal] for "[#34 Normal]" in game.

mourningsickness:
image

mourningsickness:
everything is working great, but i think the queries need some more work. the !top menus seem.. just a little bit inaccurate. i am wondering if adapting the normal rank queries didn't work perfectly. i had to rewrite one completely, and use a CAST() in it to get the points to sort as floats instead of lexicographically for some reason.

mourningsickness:
Screenshot from 2025-08-01 15-36-01

dynamic {currentstylerank} chat tag support added.

Nairdaa:
What a pogchamp.

mourningsickness:
Screenshot from 2025-08-11 18-34-27

i optimized the queries a lot, and the ranks/points are identical to before.. radical

mourningsickness:
UpdatePointsForSinglePlayer, line 1343 is giving me a SQL error when someone sets a time, but the query runs fine in a SQL editor.. so i'm not sure what the problem is:

`[shavit/shavit-rankings.smx] Timer (rankings, update all style points) error! Reason: You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'SELECT auth, style, SUM(points) as total FROM playertimes WHERE auth = 811703...' at line 1`

mourningsickness:
i got the query fixed, but think it will be for mysql only because i borrowed the JOIN from the lower query that is specified as being for mysql. i need help with these queries and testing on other db engines to find out if there are issues. as it sits, everything works great on mysql.

PR #1256, opened 2025-08-01
Group zoned maps in LoadMapZones

Awesomerly:
If a map had multiple end zones it would be in the map list multiple times. This fixes that issue.


mourningsickness:
it also improves the mapchooser to add a check for a start _and_ end zone to catch maps with missing zones

Awesomerly:
that would be good to do. not tonight though xd!

Issue #1213, opened 2024-07-21
[ Request ] perf ratio


Nairdaa:
![image](https://github.com/user-attachments/assets/c2611bee-b978-4008-875f-9fef7442e27e)
Could you please add it to the center hud too?

Same format as in the right panel: 
Jumps: 10 (%.2f)

edit: it's only 1 decimal in the right hud. Oh well. Copy pasta, please.

PR #1259, opened 2025-08-11
shavit-hud.sp - add perf% to center hud if on scroll style and HUD2_PERFS enabled

mourningsickness:
Requested here: https://github.com/shavitush/bhoptimer/issues/1213

### DIFF
--- a/addons/sourcemod/scripting/shavit-hud.sp
+++ b/addons/sourcemod/scripting/shavit-hud.sp
@@ -1369,7 +1369,15 @@ int AddHUDToBuffer_Source2013(int client, huddata_t data, char[] buffer, int max
 
 		if((gI_HUD2Settings[client] & HUD2_JUMPS) == 0)
 		{
-			FormatEx(sLine, 128, "%T: %d", "HudJumpsText", client, data.iJumps);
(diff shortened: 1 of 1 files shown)


mourningsickness:
image
works. this also made me notice there are two options in !hud for "Sync", one being the center text and the other being on the right.. but there are no indications of which is which or any sort of menu styling for the hud elements.. the whole !hud menu should probably be redone to be nicer. in addition to a separate hud setting for this? idk honestly could go either way. works as is and i doubt anyone will even notice because who plays scroll anyway?

rtldg:
> this also made me notice there are two options in !hud for "Sync", one being the center text and the other being on the right.. but there are no indications of which is which or any sort of menu styling for the hud elements.. 

Yeah, it's definitely not clear at all.

> the whole !hud menu should probably be redone to be nicer.

Probably...

> in addition to a separate hud setting for this? 

Yeah, probably should be. Do you mind adding it as another setting? It's fine in either HUD2\_ or HUD\_

rtldg:
I put it behind HUD_PERFS_CENTER, which isn't enabled by default 😇

PR #1261, opened 2025-08-28
Add new convar to spawn player inside start zone

xhaooy:
Add new convar to spawn player inside start zone and create new function to handle that (because i does same thing as shavit_zones_teleporttostart).


rtldg:
We have [`shavit_misc_startonspawn` in shavit-misc.sp](https://github.com/shavitush/bhoptimer/blob/21a7b58c82c98136fbdd43c13ba33090241642f5/addons/sourcemod/scripting/shavit-misc.sp#L275C33-L275C57). Is `shavit_zones_teleporttostart_onspawn` needed?

PR #1281, opened 2026-04-23
Update shavit-checkpoints.sp to add pausing, among other things

2x74:
+ Pausing when teleporting to a checkpoint
- Checkpoint limit
- The weird velocity loss bug that would sometimes happen when you TP to the last checkpoint

### DIFF
--- a/addons/sourcemod/scripting/shavit-checkpoints.sp
+++ b/addons/sourcemod/scripting/shavit-checkpoints.sp
@@ -1,9 +1,13 @@
 /*
- * shavit's Timer - Checkpoints
  * by: shavit, kidfearless, Nairda, GAMMA CASE, rumour, rtldg, sh4hrazad, Ciallo-Ani, olivia, Nuko, yupi2
+ * fork by: luna
+ *
+ * Changes from upstream:
+ *   - Segmented checkpoint limit is now configurable with no hard cap (0 = unlimited)
+ *   - Pause-on-teleport toggle in seg menu, unpauses on any input
+ *   - Fixed velocity being lost on the first teleport of a session
  *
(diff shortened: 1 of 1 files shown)


mourningsickness:
you should re-fork the timer and make the changes again without all of the additional stuff. just fixes applied, not comments removed, other tweaks done, etc. this is very bloated with stuff that shouldn't be in the pr

rtldg:
- This is very buggy and doesn't work properly. With a !seg checkpoint, holding +jump causes you to fly through the air, without your timer ticking, and allows you to hit the endzone and get a time.
  I didn't test any more because idc.
- I don't think `shavit_checkpoints_maxcp_seg` needs the ability to be unlimited.
  I'll push a commit shortly (attributed to you) that raises the limit to 10000 (matching `shavit_checkpoints_maxcp`) though.
- If you make a PR, it'd be best to not use your fork's `master`/main branch because if I push changes to the PR then I'm editing your master branch and it makes it a hassle for you to cleanly sync things.
  And, even if everything else worked perfectly and I had no nits, I still have to push changes (or do some more branching myself) because your PR includes changes to top-most comment & some urls.
- You removed a lot of comments that didn't need to be removed (some were fine to remove).
  This is why I took so long for me to do something with this PR: I'd open it up, see a ton of superfluous changes, get annoyed, and close it (the tab).

For the main functionality of the commit (pausing after teleporting to a checkpoint): I don't think it's a good fit for bhoptimer, in bhoptimer's current state.
The checkpoint system is already imperfect & buggy in a few ways, and adding the pausing sounds like bringing in more bugs (which this PR did!).
Plus, the current issues with *not* pausing are pretty straightforward to mitigate by just spamming teleport.

PR #1284, opened 2026-05-30
Update sql-create-tables-and-migrations.sp

Nairdaa:
sql inject found, fix pls yo


mourningsickness:
how is this injectable? all of the variables are cvars or the map/table name..

rtldg:
If you ever find a real map that exploits this on a server that hasn't ran this migration yet (pre 2018 bhoptimer versions), I'll fix it.

Nairdaa:
cba creating one, no longer around. problem "solved", leave this breach as it is

PR #1289, opened 2026-07-13
shavit-wr.sp - add player's PB to the !wr menu (like how !style menu is)

mourningsickness:
Screenshot_2026-07-12_22-20-33


rtldg:
I didn't test but it compiles so surely it's fine :)

PR #1290, opened 2026-07-22
shavit-wr.sp - show correct PB when viewing a different map

akanora:
Shavit_GetClientPB only stores times for the active server map, so !wr on any other map either hid the PB entirely or showed the current map's time. PBs are now queried from the database per map alongside WR times.


rtldg:
@mourningsickness said it works, and hopefully it does with sqlite too, and hopefully they're not a liar!!!!!

PR #1291, opened 2026-08-19
Add Replay_Automatic type

mariokeks:
Adding Replay_Automatic as replay type mainly for native usage but could be extended to the replay type selection aswell.
The basic idea was that you dont have to rely on Replay_Dynamic when using the StartReplay natives.


rtldg:
Looks rightish but I didn't test

PR #1292, opened 2026-08-21
Add Shavit_StopReplay & Shavit_GetClientReplayBot natives

mariokeks:
There was previously no clean way to stop a replay bot:
- For `Replay_Central` bots, there was no way to stop one at all. Forcibly kicking it would leave it missing for up to several seconds until the cron timer respawns it.
- For `Replay_Dynamic` bots, the only option was `KickClientEx` to kick the bot in the same frame.
- For `Replay_Prop` bots, the only option was RemoveEdict, which destroys the prop and bypasses `Shavit_OnReplayEnd`. `OnEntityDestroyed`, which handles `Replay_Prop`s that mysteriously die, only kicks them (`KickReplay`) rather than finishing them (`FinishReplay`)
```
/**
 * Stops the playback of a replay bot.
 *
 * @param bot                       The replay bot entity.
 * @param client                    Client requesting the stop, or 0 to bypass the permission check.
 * @return                          True if the replay bot has been stopped, false otherwise.
 */
native bool Shavit_StopReplay(int bot, int client = 0);
```

- Either a simple check if a client has a active replay bot or a way to get further information of the active replay via the replay bot entity.
```
/**
 * Retrieves the replay bot (or prop) entity currently tied to a client.
 *
 * @param client                    Client index.
 * @return                          Replay bot entity, 0 if no replay bot is active.
 */
native int Shavit_GetClientReplayBot(int client);
```

### DIFF
--- a/addons/sourcemod/scripting/include/shavit/replay-playback.inc
+++ b/addons/sourcemod/scripting/include/shavit/replay-playback.inc
@@ -333,7 +333,7 @@ native void Shavit_SetReplayCacheName(int bot, char[] name);
  * @param client                    Client requesting the stop, or 0 to bypass the permission check.
  * @return                          True if the replay bot has been stopped, false otherwise.
  */
-native void Shavit_StopReplay(int bot, int client = 0);
+native bool Shavit_StopReplay(int bot, int client = 0);
 
 /**
  * Starts a replay given a style and track.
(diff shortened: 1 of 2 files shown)


rtldg:
Looks rightish but I didn't test

